PR #61, opened 2024-03-06
Streaming (WIP)

elijahbenizzy:


elijahbenizzy:
OK, basic POC is working. To get this to production, we need:
- [ ] Docs
    - [ ] Overview of streaming
    - [ ] Reference API
- [ ] Demo
    - [ ] Add gpt-streaming
- [ ] Features
    - [ ] Get halt_before to work -- this has the same behavior
        - [ ] `StreamingResultContainer.is_empty` -- this allows `halt_before`
    - [ ] Get the stream call to work with non-streaming results
- [ ] Polish
    - [ ] Break into commits
    - [ ] Add docstrings for all new methods
    - [ ]  Change `StreamingResultContainer` to `StreamOutput` - feels cleaner/less stuffy
    - [ ] Implement this for consistency: https://github.com/DAGWorks-Inc/burr/issues/63
- [ ] Testing
    - [ ] Add another E2E test for `halt_before`/`halt_after`
    - [ ] Test it out with streamlit
- [ ] Async generators
    - [ ] Implement
    - [ ] Test

burr/core/application.py
```diff
@@ -584,6 +606,224 @@ async def arun(
...
+            )...build()
+            prompt = "Count to 100, with a comma between each number and no newlines. E.g., 1, 2, 3, ..."
+            action, streaming_result = application.stream_result(halt_after='streaming_response', inputs={"prompt": prompt})
+            color = "red" if action.name == "error" else "green"
+            if action.name == "streaming_response": # can also use the ``.streaming`` attribute of action
+                for result in output:
+                     print(format(result['response'], color)) # assumes that error and streaming_response both have the same output shape
+            else:
+                result, state = output.get()
+                print(format(result['response'], color))
```
skrawcz, line 721: what's the difference between these two examples?

Shouldn't `stream_result` always return a tuple?

tests/core/test_action.py
```diff
@@ -376,3 +378,112 @@ def correct_signature(state: State) -> Tuple[dict, State]:
...
+
+
+def test_streaming_action_stream_run():
+    class SimpleStreamingAction(StreamingAction):
+        def stream_run(self, state: State, **run_kwargs) -> Generator[dict, None, dict]:
+            buffer = []
+            for char in state["echo"]:
+                yield {"response": char}
+                buffer.append(char)
+            return {"response": "".join(buffer)}
```
skrawcz, line 390: shouldn't this return state too?

tests/core/test_application.py
```diff
@@ -254,11 +265,56 @@ def inputs(self) -> list[str]:
...
+class StreamingCounter(StreamingAction):
+    def stream_run(self, state: State, **run_kwargs) -> Generator[dict, None, dict]:
+        if "steps_per_couunt" in run_kwargs:
+            steps_per_count = run_kwargs["granularity"]
+        else:
+            steps_per_count = 10
+        count = state["count"]
+        for i in range(steps_per_count):
+            yield {"count": count + ((i + 1) / 10)}
+        return {"count": count + 1}
```
skrawcz, line 304: again why the no new state?

docs/concepts/streaming-actions.rst
```diff
@@ -0,0 +1,91 @@
...
+                temperature=0,
+                stream=True,
+            )
+            buffer = []
+            for chunk in response:
+                delta = chunk.choices[0].delta.content
+                buffer.append(delta)
+                yield {'response': delta}
+            full_response = ''.join(buffer)
+            return {'response': full_response}
```
skrawcz, line 45: again, why no state return? It says it writes to it below, but that isn't true. Otherwise this should be an error condition, no?

skrawcz: Oooooooh I see -- this is the Object oriented approach.

tests/core/test_action.py
skrawcz: nvm. this is the OO API -- and update() takes care of that.

examples/streaming_chatbot/application.py
```diff
@@ -0,0 +1,187 @@
...
+            "role": "assistant",
+        }
+    }
+    return result, state.update(**result).append(chat_history=result["response"])
+
+
+@streaming_action(reads=["prompt", "chat_history", "mode"], writes=["response"])
+def chat_response(
+    state: State, prepend_prompt: str, model: str = "gpt-3.5-turbo"
+) -> Tuple[dict, State]:
```
skrawcz, line 77: this has the wrong annotation -- shouldn't we error here?

tests/core/test_application.py
skrawcz: this is the OO API -- so update takes care of it.

examples/streaming_chatbot/application.py
elijahbenizzy: Not compiling against annotations now -- its just a type-hint for the user (and the IDE should be smart enough to know)

elijahbenizzy: Fixed the annotation

skrawcz (approved):

PR #65, opened 2024-03-11
Adds links to the left hand index

skrawcz:
So that it's clearer where burr lives and
links to other things.


skrawcz:
@elijahbenizzy do docs auto build? or?

PR #68, opened 2024-03-12
In-app Chatbot demo

elijahbenizzy:


elijahbenizzy:
Rough sketch. 
image

Required for production:
- [ ] Get the image generation to save to an s3 bucket behind a CDN for demos (should be able to reuse image-telephone? Create a new one?). Or just save locally... And use in static assets. Probably cleaner/easier.
- [ ] Add interactivity
    - [ ] Test out interactive endpoint
    - [ ] Add the ability to press `Send`
    - [ ] Rename `Send` to have the classic paper airplane emoji that everyone likes
    - [ ] Get it to show the full chat history on click + prompt user clicked in
    - [ ] Add ability to create new one
        - [ ] Enusre it doesn't break
- [ ] Fix hacks
    - [ ] Ensure we can import without openAI/create a client
    - [ ] Fix classpath so it just works when running
    - [ ] Add it so users know what to do if they have to create a client
        - [ ] This should be a fixed message and a specific step -- maybe add `Error`?
- [ ] More demo
    - [ ] Highlight the chat history we're currently on
    - [ ] Use that (maybe) to set the highlight/coloring

elijahbenizzy:
Bypassed this one: https://github.com/DAGWorks-Inc/burr/issues/69

skrawcz (approved):
seems fine

elijahbenizzy:
Blocked on #73 -- this leverages that so I'm going to use it.

PR #73, opened 2024-03-15
Formally adds state persistence

skrawcz:
This does a few things:

1. adds interface for persister and SQLLite implementation.
2. pushes app_id, and partition_key through to App and hooks for (1)
3. moves sequence ID incrementing to the start of a step to avoid bad restarts; we were only incrementing after saving which was bad.
4. shows API in counter example with changes to builder required to facilitate the above.
5. Adds basic docs (with TODOs to fill them out)
6. Stubs out modifying the local tracking client -- main question here is does it handle partition key or not?


burr/core/application.py
```diff
@@ -31,7 +32,14 @@
     create_action,
     default,
 )
-from burr.core.state import State
+from burr.core.state import BasicStatePersistence, State
+from burr.lifecycle import (
```
elijahbenizzy, line 36: You shouldn't need to import these. We should:
1. expose `does_hook(hook_name, is_async)` from `_does_hook` on `LifecycleAdapterSet`
2. Call `does_hook` below when we check `isinstance`

Two reasons:
1. dependencies are circular (already) -- we use `if TYPE_CHECKING` in `lifecycle.internal`. This is necessary, but best to minimize
2. `isinstance` is iffy -- it moves the burden of deciding what an object does from the object to the caller.

burr/core/application.py
```diff
@@ -235,8 +243,14 @@ def __init__(
         transitions: List[Transition],
         state: State,
         initial_step: str,
+        partition_key: str,
```
elijahbenizzy, line 246: Let's make it `optional` -- we want to allow *not* using a partition key.

burr/core/application.py
```diff
@@ -268,12 +282,11 @@ def step(self, inputs: Optional[Dict[str, Any]] = None) -> Optional[Tuple[Action
         :param inputs: Inputs to the action -- this is if this action requires an input that is passed in from the outside world
         :return: Tuple[Function, dict, State] -- the function that was just ran, the result of running it, and the new state
         """
-
-        try:
-            out = self._step(inputs=inputs, _run_hooks=True)
-            return out
-        finally:
-            self._increment_sequence_id()
+        # we need to increment the sequence before we start computing
```
elijahbenizzy, line 285: Does this break the UI or another part of the model? The removed part was done specifically so:

1. The sequence ID was always the current step being executed. Every call at the step would be 
2. It was always incremented right before the next one
3. We didn't have to start at `-1`, so the sequence ID would start at `0`

So, it would start at 0, and every step would increment it at the *very end* so the last one would start it. I don't mind how it works, but my only requirements are that (a) it starts at zero and (b) it gets incremented regardless of failure and (c) we only increment it once per `step()/_step()` call.

elijahbenizzy: Also why would it get stuck? That doesn't follow from incrementing beforehand...

burr/core/application.py
```diff
@@ -730,7 +751,7 @@ def streaming_response(state: State, prompt: str) -> Generator[dict, None, Tuple
             raise ValueError(
                 f"Cannot stream result -- no next action found! Prior action was: {self._state[PRIOR_STEP]}"
             )
-
+        self._increment_sequence_id()  # TODO: is this in the right place?
```
elijahbenizzy, line 754: So this is the right place if we move incrementing sequence ID to before (this is a makeshift step) -- this does simplify things.

burr/core/application.py
```diff
@@ -980,10 +1000,19 @@ def sequence_id(self) -> Optional[int]:
 
         :return:
         """
-        return self._state.get(SEQUENCE_ID, 0)
+        return self._sequence_id
 
     def _increment_sequence_id(self):
-        self._state = self._state.update(**{SEQUENCE_ID: self.sequence_id + 1})
+        self._sequence_id = self._sequence_id + 1
```
elijahbenizzy, line 1006: This is weird to me -- why are we keeping track of it in two places? We have `self._state` as a source of truth, we shouldn't have it as a field as well and keep it synced.

burr/core/application.py
```diff
@@ -980,10 +1000,19 @@ def sequence_id(self) -> Optional[int]:
...
+        self._state = self._state.update(**{SEQUENCE_ID: self.sequence_id})
+
+    @property
+    def id(self) -> str:
```
elijahbenizzy, line 1010: Hmm `id` feels a little non-descript. Maybe `app_id`? `uid`?

burr/core/application.py
```diff
@@ -980,10 +1000,19 @@ def sequence_id(self) -> Optional[int]:
...
+        return self._id
+
+    @property
+    def partition_key(self) -> str:
```
elijahbenizzy, line 1014: Let's add docstrings for these properties and make sure they're part of the docs.

burr/core/application.py
```diff
@@ -1041,6 +1078,23 @@ def __init__(self):
         self.actions: Optional[List[Action]] = None
         self.start: Optional[str] = None
         self.lifecycle_adapters: List[LifecycleAdapter] = list()
+        self.app_id: str = str(uuid.uuid4())
```
elijahbenizzy, line 1081: Put this in `build`, not `__init__` -- this restricts a single builder to a single `app`, which is not required. Also its weird to set it to a value that has meaning then overwrite.

burr/core/application.py
```diff
@@ -1041,6 +1078,23 @@ def __init__(self):
...
+        self.partition_key: str = None
+        self.sequence_id: int = 0
+        self.persister = None
+        self.use_entrypoint_from_save_state: bool = None
+        self.default_state: dict = None
+
+    def with_identifiers(
```
elijahbenizzy, line 1088: Full docstring, This is user-facing

burr/core/application.py
```diff
@@ -1127,38 +1181,152 @@ def with_hooks(self, *adapters: LifecycleAdapter) -> "ApplicationBuilder":
...
+        :param tracker: Tracker to use. ``local`` creates one, else pass one in.
+        :param project: Project name -- used if the tracker is local.
+        :param params: Parameters to pass to the tracker if it's local.
         :return: The application builder for future chaining.
         """
         if params is None:
             params = {}
-        if tracker == "local":
+        # if it's a lifecycle adapter, just add it
+        if isinstance(
```
elijahbenizzy, line 1202: See note above... Now that I'm looking at it, we should have a function that tells us whether something is a hook so we don't have to do this. This feels a bit ugly -- there's an implicit contract here that we don't make explicit. 

Also, why these specific ones? Have a comment/explanation.

burr/core/application.py
elijahbenizzy: On second thought you'll want to be a little more flexible -- see the note below.

burr/core/application.py
```diff
@@ -1127,38 +1181,152 @@ def with_hooks(self, *adapters: LifecycleAdapter) -> "ApplicationBuilder":
...
     def with_tracker(
-        self, project: str, tracker: Literal["local"] = "local", params: Dict[str, Any] = None
+        self,
+        tracker: Union[Literal["local"], object] = "local",
+        project: str = "default",
+        params: Dict[str, Any] = None,
     ):
         """Adds a "tracker" to the application. The tracker specifies
         a project name (used for disambiguating groups of tracers), and plugs into the
         Burr UI. Currently the only supported tracker is local, which takes in the params
```
elijahbenizzy, line 1191: Mention that there are two APIs, or kill the string-based one.

burr/core/application.py
```diff
@@ -1041,6 +1078,23 @@ def __init__(self):
...
+        self.default_state: dict = None
+
+    def with_identifiers(
+        self, app_id: str = None, partition_key: str = None, sequence_id: int = None
+    ) -> "ApplicationBuilder":
+        if app_id is not None:
+            self.app_id = app_id
+        if partition_key is not None:
+            self.partition_key = partition_key
+        if sequence_id is not None:
```
elijahbenizzy, line 1095: This is not an identifier? This is a specific aspect of the state -- I don't really see why a user would be setting it? Either:

1. The persister would be
2. It would start at zero

burr/core/application.py
```diff
@@ -1127,38 +1181,152 @@ def with_hooks(self, *adapters: LifecycleAdapter) -> "ApplicationBuilder":
...
 
-            self.lifecycle_adapters.append(LocalTrackingClient(project=project, **params))
+            kwargs = {"project": project, "app_id": self.app_id}
+            kwargs.update(params)
+            self.lifecycle_adapters.append(LocalTrackingClient(**kwargs))
+        else:
+            raise ValueError(f"Tracker {tracker}:{project} not supported")
+        return self
+
+    def initialize_from(
```
elijahbenizzy, line 1223: docstring

burr/core/application.py
```diff
@@ -1127,38 +1181,152 @@ def with_hooks(self, *adapters: LifecycleAdapter) -> "ApplicationBuilder":
...
+        self.persister = persister
+        self.resume_at_next_action = resume_at_next_action
+        self.default_state = default_state
+        self.start = default_entry_point
+        return self
+
+    def with_state_persister(
+        self, persister: BasicStatePersistence, on_every: str = "step"
+    ) -> "ApplicationBuilder":
+        """Adds a state persister to the application. This is a way to persist state out to a database, file, etc...
```
elijahbenizzy, line 1239: More info about how this works

burr/core/application.py
```diff
@@ -1127,38 +1181,152 @@ def with_hooks(self, *adapters: LifecycleAdapter) -> "ApplicationBuilder":
...
+        if on_every != "step":
+            raise ValueError(f"on_every {on_every} not supported")
+        # stick persister into a post_step_hook if it doesn't implement it already.
+        if hasattr(persister, "post_step_hook"):
+            if persister not in self.lifecycle_adapters:
+                self.lifecycle_adapters.append(persister)
         else:
-            raise ValueError(f"Tracker {tracker} not supported")
+            # TODO: pull this out more formally
+            class persisterWrapper(PostRunStepHook):
```
elijahbenizzy, line 1254: Class capitalization?

elijahbenizzy: Define this elsewhere -- this is muddying up the application , which is already way too long. Probably should live with the persister.

burr/core/application.py
```diff
@@ -1127,38 +1181,152 @@ def with_hooks(self, *adapters: LifecycleAdapter) -> "ApplicationBuilder":
...
+         - self.sequence_id
+         - maybe self.start
+
+        """
+        # load state from persister
+        load_result = self.persister.load(self.partition_key, self.app_id, self.sequence_id)
+        if load_result is None:
+            # there was nothing to load -- use default state
+            self.state = self.state.update(**self.default_state)
+            self.sequence_id = 0
```
elijahbenizzy, line 1292: This will get it out of sync with the `increment_sequence_id` call... We should just have the `sequence_id` be a property that gets the state that's never set directly, and have a getter on it

burr/core/application.py
```diff
@@ -1127,38 +1181,152 @@ def with_hooks(self, *adapters: LifecycleAdapter) -> "ApplicationBuilder":
...
+        else:
+            # there was something
+            last_position = load_result["position"]
```
elijahbenizzy, line 1295: IMO dataclasses might be slightly nicer than `TypedDict`, but don't weight this too heavily

burr/core/application.py
```diff
@@ -1127,38 +1181,152 @@ def with_hooks(self, *adapters: LifecycleAdapter) -> "ApplicationBuilder":
...
+            self.state = load_result["state"]
+            self.sequence_id = load_result["sequence_id"]
+            status = load_result["status"]
+            if self.resume_at_next_action:
+                # if we're supposed to resume where we saved from
+                if status == "completed":
```
elijahbenizzy, line 1301: The more I'm looking through this the more I think that state persistence should *not* know about whether it errored. How will it change which state to load? Do we even use it? It just sticks a ton of complexity here.

elijahbenizzy: Also I don't think we should change `self.start` -- then we forget where we *should* start from, right? That's a fixed property, not something that's dynamic. We use `previous_step` + might need some ability to reset.

burr/core/application.py
```diff
@@ -1127,38 +1181,152 @@ def with_hooks(self, *adapters: LifecycleAdapter) -> "ApplicationBuilder":
...
+        at the specified interval.
+
+        :param persister: The persister to add
+        :param on_every: The interval to persist state. Currently only "step" is supported.
+        :return: The application builder for future chaining.
+        """
+        if on_every != "step":
+            raise ValueError(f"on_every {on_every} not supported")
+        # stick persister into a post_step_hook if it doesn't implement it already.
+        if hasattr(persister, "post_step_hook"):
```
elijahbenizzy, line 1249: This is really confusing me -- we have `initialize()` as a method here, but this sets persister, but initialize may or may not have the same object?

I think we should plan for just one use-case, write and read from the same thing. Otherwise its purely implemented using hooks (and maybe a wrapper that turns some things off).

burr/core/state.py
```diff
@@ -196,3 +198,231 @@ def __iter__(self) -> Iterator[Any]:
...
+    partition_key: str
+    app_id: str
+    sequence_id: int
+    position: str
+    state: State
+    created_at: str
+    status: str
+
+
+class BasicStatePersistence(abc.ABC):
```
elijahbenizzy, line 213: Persister, not persistence

burr/core/state.py
```diff
@@ -196,3 +198,231 @@ def __iter__(self) -> Iterator[Any]:
...
+        :param app_id:
+        :param sequence_id:
+        :param position:
+        :param state:
+        :param status:
+        """
+        pass
+
+
+class SQLLitePersistence(BasicStatePersistence):
```
elijahbenizzy, line 270: Add `TODO  -- use ORM` -- this codes going to get a lot easier to debug IMO.

burr/tracking/client.py
```diff
@@ -237,6 +244,32 @@ def post_end_span(
     def __del__(self):
         self.f.close()
 
+    def initialize(self):
+        # not needed because this one does things in the constructor
```
elijahbenizzy, line 248: Remove from constructor then?

docs/concepts/state-persistence.rst
```diff
@@ -0,0 +1,24 @@
...
+Key to writing a real life `burr` application is state persistence. E.g. you're building a chat bot and you
+want to store the conversation history and then reload it when you restart. Or, you have a long running process,
+or series of agents, and you want to store the state of the process after each action, and then reload it if it fails, etc.
+
+Here we'll walk through how to add state persistence to a `burr` application. The following walk through the relevant
+ApplicationBuilder() methods, and then a full toy example.
+
+.initialize_from() method
+_________________________
+TODO:
```
elijahbenizzy, line 16: This should get done before merging

docs/examples/ml_training.rst
```diff
@@ -2,4 +2,9 @@
 ML Model Training
 ====================
 
+Hyperparameter Optimization is a key part of training machine learning models. For more sophisticated examples that
+aren't grid search, `burr` is a quick and easy way to implement hyperparameter optimization. You want to keep
```
elijahbenizzy, line 6: It's probably just as good for grid search TBH. Also, no parallelism. This is a good reason to make parallelism/multiple async steps a first-class citizen later on.

docs/main.rst
```diff
@@ -4,6 +4,8 @@ Burr
 
 Welcome to Burr's documentation, the state machine for managing complex data/LLM projects.
 
+TODO: VIDEO.
```
elijahbenizzy, line 7: Only add a TODO if you're going to do before merging, otherwise add an issue.

docs/reference/persister.rst
```diff
@@ -0,0 +1,21 @@
...
+This is an interface one should implement to provide state persistence for your
+application.
+
+.. autoclass:: burr.core.state.BasicStatePersistence
+   :members:
+
+   .. automethod:: __init__
+
+
+Supported Implementations
```
elijahbenizzy, line 14: Add tracker with a note about why we're doing it this way

examples/counter/application.py
```diff
@@ -1,41 +1,61 @@
...
+from burr.core.state import SQLLitePersistence
 from burr.lifecycle import LifecycleAdapter
 
+logger = logging.getLogger(__name__)
+
 
 @action(reads=["counter"], writes=["counter"])
 def counter(state: State) -> Tuple[dict, State]:
     result = {"counter": state["counter"] + 1}
+    # import random
```
elijahbenizzy, line 16: delete

examples/multi-agent-collaboration/hamilton_application.py
```diff
@@ -1,17 +1,66 @@
 import json
+from typing import Any, Dict, Optional
 
 import func_agent
-from hamilton import driver
+from hamilton import driver, lifecycle
```
elijahbenizzy, line 5: `import lifecycle as h_lifecycle`

examples/multi-agent-collaboration/hamilton_application.py
```diff
@@ -1,17 +1,66 @@
...
+        success: bool,
+        task_id: Optional[str],
+        run_id: str,
+        **future_kwargs: Any,
+    ):
+        context_manager = self.active_spans.pop(node_name)
+        context_manager.__exit__(None, None, None)
+
+
+def initialize_tool_dag(agent_name: str, tracer: TracerFactory) -> driver.Driver:
```
elijahbenizzy, line 57: What does `tool` mean here? Are we using elsewhere?

elijahbenizzy (changes requested):
OK, big things:
1. Persistence handling failure is confusing me quite a bit, and we don't actually use it
2. Sequence IDs are going to start at 1 (unless I'm mistaken)
3. API is confusing me -- maybe docs will clear this up? We have `trackers`, `persisters`, `initialize_from`, etc... and its far from obvious which does which. Might need to revisit.
4. Docs, remove a few TODOs prior to merge

burr/core/application.py
elijahbenizzy: Answer -- we start at the beginning and it all gets simpler

PR #74, opened 2024-03-17
MVP for telemetry

elijahbenizzy:
This just:
1. Tracks fn usage
2. Hides any user information

This is not particularly in-depth, but it's about what we need now to see if people are picking this up.


docs/reference/telemetry.rst
```diff
@@ -0,0 +1,47 @@
...
+    telemetry.disable_telemetry()
+
+2. Set the key telemetry_enabled to false in ``~/.hamilton.conf`` under the DEFAULT section:
+
+.. code-block:: ini
+
+    [DEFAULT]
+    telemetry_enabled = False
+
+3. Set HAMILTON_TELEMETRY_ENABLED=false as an environment variable. Either setting it for your shell session:
```
skrawcz, line 37: copy paste error

skrawcz (approved):

PR #87, opened 2024-03-20
Adds postgresql persister

skrawcz:
This is a simple implementation that uses JSON
to serialize the state to. Very similar in spirit to the SQLLITE one, but using postgresql constructs.

I set varchar limits so they wouldn't be text.
These obviously could be tightened.

### DIFF
--- a/burr/integrations/persisters/postgresql.py
+++ b/burr/integrations/persisters/postgresql.py
@@ -1,0 +1,234 @@
+try:
+    import psycopg2
+except ImportError:
+    raise ImportError(
+        "Missing plugin PostgreSQL! To use the PostgreSQL plugin, you must install the following libraries: psycopg2. "
+        "You can install this with burr[postgresql] or pip install psycopg2 (replace with your package manager of choice)."
+    )
+
+import json
+import logging
+from typing import Literal, Optional
+
+from burr.core import persistence, state
+
+logger = logging.getLogger(__name__)
+
+
+class PostgreSQLPersister(persistence.BaseStatePersister):
+    """Class for PostgreSQL persistence of state. This is a simple implementation.
+
+    To try it out locally with docker -- here's a command -- change the values as appropriate.
+
+    .. code:: bash
+
+        docker run --name local-psql \  # container name
+                   -v local_psql_data:/SOME/FILE_PATH/ \  # mounting a volume for data persistence
+                   -p 54320:5432 \  # port mapping
+                   -e POSTGRES_PASSWORD=my_password \  # superuser password
+                   -d postgres  # database name
+
+    Then you should be able to create the class like this:
+
+    .. code:: python
+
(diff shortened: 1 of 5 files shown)


elijahbenizzy:
There's actually no real difference between varchar and text, postgres treats them the same. See https://stackoverflow.com/questions/4848964/difference-between-text-and-varchar-character-varying. 

Otherwise, looks good. Let's add a version/schema?

elijahbenizzy (commented):
See comment in PR about varchar, and schema version. We should have some way that we can update it.

skrawcz:
> See comment in PR about varchar, and schema version. We should have some way that we can update it.

I actually had it text before I changed it 😶‍🌫️ ... schema version would only make sense for the state JSON schema.

elijahbenizzy:
Also, let's change the persister API to require `**kwargs` so we can add more of them and make it backwards compatible

burr/integrations/persisters/postgresql.py
```diff
@@ -0,0 +1,234 @@
+try:
```
elijahbenizzy, line 1: Let's stick this in integrations actually? Or at least use the same pattern here: https://github.com/DAGWorks-Inc/burr/blob/main/burr/tracking/client.py#L30. It automates away the message so they're all the same.

burr/integrations/persisters/postgresql.py
```diff
@@ -0,0 +1,234 @@
...
+    .. code:: python
+
+        p = PostgreSQLPersister.from_values("postgres", "postgres", "my_password",
+                                           "localhost", 54320, table_name="burr_state")
+
+
+    """
+
+    @classmethod
+    def from_config(cls, config: dict) -> "PostgreSQLPersister":
```
elijahbenizzy, line 42: Why do we have this? Just postgres convention?

burr/integrations/persisters/postgresql.py
```diff
@@ -0,0 +1,234 @@
...
+                   -v local_psql_data:/SOME/FILE_PATH/ \  # mounting a volume for data persistence
+                   -p 54320:5432 \  # port mapping
+                   -e POSTGRES_PASSWORD=my_password \  # superuser password
+                   -d postgres  # database name
+
+    Then you should be able to create the class like this:
+
+    .. code:: python
+
+        p = PostgreSQLPersister.from_values("postgres", "postgres", "my_password",
```
elijahbenizzy, line 35: persistor, not `p` -- var names

burr/integrations/persisters/postgresql.py
skrawcz: I'd want it.

burr/integrations/persisters/postgresql.py
skrawcz: it is in integrations, but under persisters.

elijahbenizzy (approved):

PR #90, opened 2024-03-21
Version bump 0.9.0

elijahbenizzy:


elijahbenizzy:
Release every feature at first, then release week by week once we hit `1.0.0`

PR #115, opened 2024-03-30
Adds CLI command to help generate test cases

skrawcz:
The idea is that someone is in the UI and wants to fix/create
a test around some part of the application.

They can then use the command line to bootstrap it. We could
do this all in the UI too eventually.

We assume they want to then have a JSON file with input state
and expected output state to validate against. I think this
is a nice way to do it. Then it's all checked in and they can run
it in CI easily...

This command assumes a local tracker for now. In the future
we can add a yaml/config file that can instantiate a persister
to load data instead.

Otherwise we assume pytest, we could create a unittest stub too,
and also that users will like the file drive approach.

See notebook for example usage.


docs/examples/creating_tests.rst
```diff
@@ -0,0 +1,14 @@
+====================
+Creating Test Cases
+====================
+
+With Burr tracking state is part of the framework. This means creating a realistic test case
+for an action involves turning a persister or tracker on and then pulling that state out
+for a test case. The following example demonstrates how to create a test case
+using the `burr-test-case` command.
```
skrawcz, line 8:
```suggestion
With Burr tracking state is part of the framework. This means creating a realistic test case
for an action involves turning a persister or tracker on and then pulling that state out
for a test case. The following example demonstrates how to create a test case
using the `burr-test-case` command. See `burr-test-case --help` for details.
```

burr/cli/__main__.py
```diff
@@ -175,10 +177,113 @@ def generate_demo_data():
...
+@click.option("--project-name", required=True, help="Name of the project")
+@click.option("--partition-key", required=True, help="Partition key to look at")
+@click.option("--app-id", required=True, help="App ID to pull from.")
+@click.option("--sequence-id", required=True, help="Sequence ID to pull.")
+@click.option(
+    "--target-file-name",
+    required=False,
+    help="What file to write the data to. Else print to console.",
+)
+def create_test_case(
```
elijahbenizzy, line 226: We should note that this is `local` -- either as a subcommand or a parameter.

Also maybe add a `testing` file in the CLI and import it? This is getting a little unwieldy.

elijahbenizzy (commented):
Looks good -- I think we can reposition a bit to capture a more core LLM case

burr/cli/__main__.py
```diff
@@ -175,10 +177,113 @@ def generate_demo_data():
...
+        else:
+            current_testcases = [tc_json]
+        print(f"\nWriting data to file {target_file_name}")
+        with open(target_file_name, "w") as f:
+            json.dump(current_testcases, f, indent=2)
+    else:
+        logger.info(json.dumps(tc_json, indent=2))
+    # print out python test to add
+    print("\nAdd the following to your test file:\n")
+    print(PYTEST_TEMPLATE.format(FILE_NAME=target_file_name, ACTION_NAME=data["position"]))
```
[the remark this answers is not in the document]
skrawcz: I think this is simpler. The way things are set up you're going to be adding to the same file, but not needing to add a new test (likely).

burr/cli/__main__.py
skrawcz: what do you mean local?  Otherwise we should just refactor on the next addition.

docs/examples/creating_tests.rst
```diff
@@ -0,0 +1,30 @@
...
+burr-test-case create  \
+  --project-name "SOME_NAME" \
+  --partition-key "SOME_KEY" \
+  --app-id "SOME_ID" \
+  --sequence-id 0 \
+  --target-file-name /tmp/test-case.json
+```
+
+See `github repository example <https://github.com/DAGWorks-Inc/burr/tree/main/examples/test-case-creation>`_
+for an example.
```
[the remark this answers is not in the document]
skrawcz: yep I think added that to the docs.

burr/cli/__main__.py
elijahbenizzy: Sorry I mean it uses the `LocalTrackingClient`. E.G. namespace now then leave room for more later.

elijahbenizzy (approved):

PR #119, opened 2024-03-31
Email assistant demo

elijahbenizzy:


examples/email-assistant/application.py
```diff
@@ -0,0 +1,174 @@
...
+        .with_actions(
+            process_input,
+            determine_clarifications,
+            clarify_instructions,
+            formulate_draft,
+            process_feedback,
+            final_result,
+        )
+        .with_transitions(
+            ("process_input", "determine_clarifications"),
```
skrawcz, line 150: these could also just be function pointers instead of strings.

examples/email-assistant/application.py
```diff
@@ -0,0 +1,174 @@
...
+            (
+                "determine_clarifications",
+                "clarify_instructions",
+                expr("len(clarification_questions) > 0"),
```
skrawcz, line 154: for later -- but these could be a function that takes in state and outputs a boolean.

elijahbenizzy: That is supported: `Condition(keys=['clarification_questions'], resolver=lambda state: len(state['clarification_questions']> 0)`

examples/email-assistant/application.py
elijahbenizzy: Yeah but its a bit easy to shoot yourself in the foot, as it depends on whether we're using the named or not-named version (E.G. `*args` from the above or `**kwargs`, or functions or action-objects).


Might be some syntactic sugar helps but we can't do the same thing we did with Hamilton as we're doing purely procedural (rather than declarative) graph-building.

skrawcz (approved):

PR #122, opened 2024-04-01
Fixes langchain message serialization

skrawcz:
Didn't find the appropriate code until now.

We can now then properly deserialize the JSON too - so the example is now updated as well.


elijahbenizzy (commented):
Looks fine for now, we'll want to revisit custom ser/deser later but not now

elijahbenizzy (approved):

PR #137, opened 2024-04-04
Fleshes out the docs for simulations

elijahbenizzy:
We need a real example but this will do for now.

https://github.com/DAGWorks-Inc/burr/issues/134


examples/ml-training/application.py
```diff
@@ -1,110 +0,0 @@
-import burr.core.application
```
skrawcz, line 1: why remove this? this stub is fine -- just need to comment what needs to be added.

skrawcz (approved):

PR #145, opened 2024-04-07
Updates telephone example

skrawcz:
To make it a top-level example.

Simplifies it a little and splits out the two implementations to make the code simpler to look at.

Adds an example hook to save images locally.

## Changes
 - examples folder for image-telephone

## How I tested this
 - locally

## Notes
 - we can have the blog post link to the simple example here too.

## Checklist


examples/image-telephone/README.md
```diff
@@ -6,17 +6,18 @@ You provide an initial image to ChatGPT, which then generates a caption. The cap
 is saved to state. That caption is then provided to DALL-E, which generates an image, which is saved to state.
 You then provide that image to ChatGPT, and so on.
 
-Right now the teriminal state is set to 2, so the game will end after 2 iterations:
```
elijahbenizzy, line 9: What are they going to learn by reading this example? Worth adding.

examples/image-telephone/application.py
```diff
@@ -0,0 +1,170 @@
...
+from burr.core import Action, ApplicationBuilder, State, default, expr
+from burr.core.action import action
+from burr.lifecycle import PostRunStepHook
+
+# import hamilton modules
+caption_images = dataflows.import_module("caption_images", "elijahbenizzy")
+generate_images = dataflows.import_module("generate_images", "elijahbenizzy")
+
+
+class ImageSaverHook(PostRunStepHook):
```
elijahbenizzy, line 23: Hmm I wonder if this should be a persister? It can save the image and load up the image URL?

examples/image-telephone/application.py
```diff
@@ -0,0 +1,170 @@
...
+            with open(os.path.join(self.path, image_name), "wb") as f:
+                f.write(requests.get(image_url).content)
+                print(f"Saved image to {self.path}/{image_name}")
+
+
+@action(
+    reads=["current_image_location"],
+    writes=["current_image_caption", "image_location_history"],
+)
+def image_caption(state: State, caption_image_driver: driver.Driver) -> tuple[dict, State]:
```
elijahbenizzy, line 51: We're not using the `Hamilton` abstraction -- this is going to confuse people. MIght be worth deleting that or fixing it up to include `overrides` if we don't like it, otherwise we should not have two ways to do things.

examples/image-telephone/application.py
```diff
@@ -0,0 +1,170 @@
...
+        .with_state(
+            current_image_location=starting_image,
+            current_image_caption="",
+            image_location_history=[],
+            image_caption_history=[],
+            caption_analysis=[],
+        )
+        .with_actions(
+            caption=image_caption.bind(caption_image_driver=caption_image_driver),
+            analysis=caption_embeddings.bind(caption_image_driver=caption_image_driver),
```
elijahbenizzy, line 129: `analysis` is confusing -- maybe call it `create_embeddings`?

examples/integrations/hamilton/README.md
```diff
@@ -0,0 +1,7 @@
+# Examples using the Hamilton integration
+
+Burr has some syntactic sugar for integrating with Hamilton. Find examples of them here.
```
elijahbenizzy, line 3: We're not using any of the syntactic sugar, might as well remove this (or fix it up to use it).

examples/image-telephone/application.py
elijahbenizzy: Ahh I see one that uses it. Let's make it clear that we have both of these.

elijahbenizzy (commented):
Looks good, two points:
1. We should make it more clear _why_ we're showing this example (the other examples are (a) something they'll want to do or (b) something they'll recognize, this is something that is just for fun, but teaches something interesting
2. We should clarify when we use the Hamilton integration and when we don't, two different paths is going to confuse people.

examples/image-telephone/application.py
skrawcz: the action is for analysis -- not part of the telephone game.

PR #146, opened 2024-04-08
Handling case when state is None

f0rodo:
I created a custom presister for Postgres and was getting this:

```python
    presister = PostgreSQLPersister()
    presister.initialize()
    return ApplicationBuilder().with_actions(
        human_input=human_input,
    ).with_transitions(

    ).initialize_from(
            initializer=presister,
            resume_at_next_action=True,
            default_state={"chat_history": []},
            default_entrypoint="human_input"
    ).with_state_persister(
        presister
    ).with_identifiers(
        app_id=f"{custormer_id}", partition_key=str(pk)
    ).build()
```

[Short description explaining the high-level reason for the pull request]

## Checklist

- [ ] PR has an informative and human-readable title (this will be pulled into the release notes)
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future TODOs are captured in comments
- [ ] Project documentation has been updated if adding/changing functionality.

PR #147, opened 2024-04-08
Adds guard for persisters returning bad state

skrawcz:
The entire call from load_state should be None if something isn't found.

Otherwise the persister should fix its assumptions/make sure None state is not saved.

## Changes
 - ApplicationBuilder code
 - test for it

## How I tested this
 - locally

## Notes
 - Spurred by discussion in #146 

## Checklist


elijahbenizzy (approved):
Looks good

PR #154, opened 2024-04-12
Adds JSON view for numeric list -- this is easier to deal with

elijahbenizzy:
We can just reuse the JSON view for this, its pretty clean.
image

### DIFF
--- a/telemetry/ui/src/components/routes/app/DataView.tsx
+++ b/telemetry/ui/src/components/routes/app/DataView.tsx
@@ -261,7 +261,7 @@ const RenderedField = (props: {
         props.value.length > 0 &&
         typeof props.value[0] === 'number' ? (
           <div key={key + '-' + String(level)}>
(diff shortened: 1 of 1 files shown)


skrawcz:
will this work for top level number arrays too? I remember having code in two spots for this reason (IIRC).

elijahbenizzy:
> will this work for top level number arrays too? I remember having code in two spots for this reason (IIRC).

From what I can tell that's actually not necessary. `FormRenderer` renders the data, but only the top level. That's always a `dict`/`object`, which means we just have to get the fields. The `RenderedField` is used recursively.

skrawcz:
Cool -- but I see you've injected a `values` field in there though.

skrawcz:
Screen Shot 2024-04-12 at 4 42 49 PM
fixed

skrawcz:
note: rendering is a little sluggish with this.

skrawcz (approved):

PR #155, opened 2024-04-12
Adds better error for missing inputs

skrawcz:
Trying to make it simpler for someone to figure out what to fix.

E.g. if they do something like:

```python
app.run(halt_after=["response"])
```
but input is required, they'll get:
```
ValueError: Action prompt is missing required inputs: {'prompt'}. Has inputs: {}.
Please provide the keyword argument `inputs` to however you're running the burr application.
e.g.
   app.run(halt_..., # your halt logic
           inputs={'prompt': 'FILL ME IN'}  # <-- this is what you're missing
   )
```

## Changes
 - core/application.py

## How I tested this
 - locally

## Checklist

### DIFF
--- a/burr/core/application.py
+++ b/burr/core/application.py
@@ -55,6 +55,14 @@ PRIOR_STEP = "__PRIOR_STEP"
 SEQUENCE_ID = "__SEQUENCE_ID"
 
 
+ERROR_MESSAGE = (
+    "-------------------------------------------------------------------\n"
+    "Oh no an error! Need help with Burr?\n"
+    "Join our discord and ask for help! https://discord.gg/4FxBMyzW5n\n"
+    "-------------------------------------------------------------------\n"
+)
+
+
 def _validate_result(result: dict, name: str) -> None:
     if not isinstance(result, dict):
         raise ValueError(
@@ -186,7 +194,8 @@ def _create_dict_string(kwargs: dict) -> str:
 
 def _format_error_message(action: Action, input_state: State, inputs: dict) -> str:
     """Formats the error string, given that we're inside an action"""
-    message = f"> Action: {action.name} encountered an error!"
+    message = ERROR_MESSAGE
+    message += f"> Action: `{action.name}` encountered an error!"
     padding = " " * (80 - len(message) - 1)
     message += padding + "<"
     message += "\n> State (at time of action):\n" + _create_dict_string(input_state.get_all())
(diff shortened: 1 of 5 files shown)


burr/core/application.py
```diff
@@ -395,11 +397,11 @@ def _process_inputs(self, inputs: Dict[str, Any], action: Action) -> Dict[str, A
...
                 processed_inputs[key] = inputs.pop(key)
-        if len(inputs) > 0:
-            raise ValueError(
+        if len(inputs) > 0 and logger.isEnabledFor(logging.DEBUG):
+            logger.debug(
                 f"Keys {inputs.keys()} were passed in as inputs to action "
-                f"{action.name}, but not declared by the action as an input! "
-                f"Action needs: {action.inputs}"
+                f"{action.name}, but not declared by the action as an input. "
+                f"Action only needs: {action.inputs}."
```
elijahbenizzy, line 404: Reword this, with the new functionality its supposed to happen. Even in debug it'll confuse people, so add more clarification.

elijahbenizzy (commented):
We need:
1. Tests -- There should be tests that would break earlier but work now
2. Documentation -- we should clearly show how this works in the concepts (https://burr.dagworks.io/concepts/actions/#runtime-inputs)

Otherwise

burr/core/application.py
```diff
@@ -1194,7 +1217,7 @@ def with_state(self, **kwargs) -> "ApplicationBuilder":
         """
         if self.initializer is not None:
             raise ValueError(
```
elijahbenizzy, line 1219: For the future, we may want to decorate these with something that catches the error and rethrows with help info, but this will do for now.

docs/concepts/actions.rst
```diff
@@ -96,87 +137,82 @@ Note that if the action has inputs, you have to define the optional ``inputs`` p
...
+If you just want to fill a result from the state, you can use the `Result` action:
 
-    @action(reads=["var_from_state"], writes=["var_to_update"])
-    def custom_action(state: State, increment_by: int) -> Tuple[dict, State]:
-        result = {"var_to_update": state["var_from_state"] + increment_by}
-        return result, state.update(**result)
+.. code-block:: python
 
     app = ApplicationBuilder().with_actions(
-        custom_action=custom_action
```
elijahbenizzy, line 143: Hmm somethig is weird with the diffing...

elijahbenizzy: Yeah its just that git diff isn't happyw ith it moving around, ignore for now.

elijahbenizzy (approved):
Looks good! Add `0.14.0` commit to this and release/bump.

PR #174, opened 2024-05-08
Handles `:` in path names (+ more for 0.15.1)

skrawcz:
`:` isn't supported by windows machines.

Therefore install doesn't work on windows.
We need to therefore:

1. update demo data
2. change tracking client to not allow `:` in the path to be present.

## Checklist


burr/tracking/client.py
```diff
@@ -85,7 +85,10 @@ def __init__(
 
     @classmethod
     def get_storage_path(cls, project, storage_dir) -> str:
-        return os.path.join(os.path.expanduser(storage_dir), project)
+        # need to replace `:` with `_` for the project_id; it becomes part of the path and characters
```
elijahbenizzy, line 88: I don't think this is the right solution. Better to error if it doesn't match. We can fix upstream, or add a patch to have `demo_` instead of `demo:`

burr/tracking/client.py
```diff
@@ -85,10 +85,7 @@ def __init__(
 
     @classmethod
     def get_storage_path(cls, project, storage_dir) -> str:
-        # need to replace `:` with `_` for the project_id; it becomes part of the path and characters
-        # need to work on windows, etc.
-        sanitized_project_id = project.replace(":", "_")
-        return os.path.join(os.path.expanduser(storage_dir), sanitized_project_id)
```
skrawcz: issue is that unbeknown to people the project name passed into the tracker gets turned into a path.  so either we enforce what a project name can be, or we still need this and instead save the project name as part of metadata and not rely on the UI to assume that the directory name maps to the project name -- that would be longer term fix.

skrawcz: I think it's a bad design assumption on our part to assume the project name is something that becomes part of the file path verbatim. If the project name has `/` in it, etc. then there's more to sanitize, etc.

elijahbenizzy: Yeah, its sloppy. As we discussed in person we'll add guards on the project name (fail earlier). 

Not too concerned (the local tracking client is not prod-ready anyway), but we want to reduce onboarding friction.

elijahbenizzy (approved):
Ship it

PR #175, opened 2024-05-11
Adds modular SERDE approach

skrawcz:
This PR introduces a modular SERDE approach.

1. Singledispatch to register type serializers.
2. Any complex object deserialization, needs to be accompanied with a "type field" that will map to the name
of a "string dispatcher function". i.e. like single dispatch, but you register a string value for it. This approach
means that most people will get around the need for field level deserialization / having to tell Burr about state types explicitly.

This skips field level serde registration for now. But leaves room for it.

Otherwise this custom serde for:

* langchain docs & messages
* pydantic objects
* some more complex serialization, e.g. pandas dataframe serialization
* ability to register classes to be pickled

TODOs:

- [x] decide on module structure and where it should live
- [x] wire changes through to tracker serialization
- [x] wire changes through to persister serialization
- [x] ~wire changes through to UI reading local tracker logs~ Not required.
- [x] update LC examples
- [x] write tests
- [x] add documentation

## How I tested this
 - unit tests
 - running lcel example

## Checklist


skrawcz:
Update: just need to test more, but I think I have a good design (and update libraries to latest locally...)

core.serde has functions.
core.state uses it and imports integrations
state.State knows how to serialize & deserialize using it.

People can then register their own for their application -- I haven't tested overriding but I assume it'll work.
Added example of pandas serializing somewhere and just showing link back...

Tracker models should be dumb -- so instead I push serialization calling into hooks.
Also persisters updated to use serialization.

Once I merge and rebase off of #215  can add persister tests that will also then check for this...

burr/core/persistence.py
```diff
@@ -140,15 +140,17 @@ def save(
 class SQLLitePersister(BaseStatePersister):
     """Class for SQLLite persistence of state. This is a simple implementation."""
 
-    def __init__(self, db_path: str, table_name: str = "burr_state"):
+    def __init__(self, db_path: str, table_name: str = "burr_state", serde_kwargs: dict = None):
```
elijahbenizzy, line 143: Hmm, a little worried about global kwargs... Won't this have the potential for clashing?

burr/core/state.py
```diff
@@ -272,3 +295,10 @@ def __iter__(self) -> Iterator[Any]:
 
     def __repr__(self):
         return self.get_all().__repr__()  # quick hack
+
+
+for serde_plugin in ["langchain", "pydantic", "pandas"]:
```
elijahbenizzy, line 300: Comment as to why this lives here

burr/integrations/__init__.py
```diff
@@ -1,34 +1,32 @@
...
-
-    print(lc_doc, lc_doc_2)
-    print(lc_message, lc_message_2)
-
-    lc_dws = _DocumentWithState(page_content="Hello, World document with state!")
-    t = serde.serialize(lc_dws)
-    print(lc_dws)
-    print(t)
-    print(serde.deserialize(t))
+# import importlib
```
elijahbenizzy, line 1: delete commented out code

burr/integrations/serde/pandas.py
```diff
@@ -0,0 +1,31 @@
+# try to import to serialize Pandas Objects
+import hashlib
+import os
+
+import pandas as pd
```
elijahbenizzy, line 5: Good to use the import guards we have above. We can add it for debugging and catch it more explicitly. Not essential though.

burr/integrations/serde/pandas.py
```diff
@@ -0,0 +1,31 @@
...
+    # Return the hexadecimal representation of the hash
+    file_name = f"df_{hash_object.hexdigest()}.parquet"
+    kwargs = pandas_kwargs.copy()
+    base_path: str = kwargs.pop("path")
+    saved_to = os.path.join(base_path, file_name)
+    value.to_parquet(path=saved_to, **kwargs)
+    return {serde.KEY: "pandas.DataFrame", "path": saved_to}
+
+
+@serde.deserializer.register("pandas.DataFrame")
```
elijahbenizzy, line 26: Do we allow overwriting serialization capabilities?

burr/integrations/serde/pandas.py
```diff
@@ -0,0 +1,31 @@
...
+def deserialize_pandas_df(value: dict, pandas_kwargs: dict, **kwargs) -> pd.DataFrame:
+    kwargs = pandas_kwargs.copy()
+    if "path" in kwargs:
```
elijahbenizzy, line 29: So there's going to be one global `path` argument? Will this work if you have more than one dataframe in state?

burr/integrations/persisters/postgresql.py
```diff
@@ -73,14 +73,19 @@ def from_values(
...
         """Constructor
 
         :param connection: the connection to the PostgreSQL database.
         :param table_name:  the table name to store things under.
         """
         self.table_name = table_name
         self.connection = connection
+        self.serde_kwargs = serde_kwargs or {}
+
+    def set_serde_kwargs(self, serde_kwargs: dict):
```
elijahbenizzy, line 86: Where is this used/what's the use-case for this?

burr/tracking/common/models.py
```diff
@@ -146,30 +146,31 @@ class BeginEntryModel(IdentifyingModel):
...
-    elif hasattr(d, "model_dump"):  # generic pydantic object
-        return d.model_dump()
-    elif hasattr(d, "to_json"):
-        return d.to_json()
-    return d
+        # return _serialize_object(_filter_inputs(inputs))
+        return _filter_inputs(inputs)
+
+
+# def _serialize_object(d: object) -> Union[dict, list, object, str]:
```
elijahbenizzy, line 153: delete

burr/tracking/server/run.py
```diff
@@ -82,6 +82,18 @@ async def ready() -> bool:
     return True
 
 
+@app.get("/api/v0/version")
```
elijahbenizzy, line 85: This is only useful for the demos, right? Otherwise `burr` will not be included, as it is not needed. 

Add a comment on use-case.

elijahbenizzy (commented):
Largely looks good, I'm confused about the serde kwargs TBH.

burr/tracking/server/run.py
skrawcz: What do you mean not included? That's how the server is installed?

burr/integrations/serde/pandas.py
skrawcz: We don't want them here as they aren't user imported. The framework tries to import them. So the message is moot.

burr/tracking/server/run.py
elijahbenizzy: In production contexts, `burr` will not be installed in the server. That said, we have no production deployments yet...

burr/integrations/serde/pandas.py
elijahbenizzy: Sure, just figured it would be nice for consistency, but you're right it is not the same.

burr/integrations/serde/pandas.py
skrawcz: whatever single dispatch does...

burr/core/persistence.py
skrawcz: no, because each entry in `serde_kwargs` would be namespaced by that serialization implementation. `pandas_kwargs` vs `pydantic_kwargs` etc.

burr/core/serde.py
```diff
@@ -0,0 +1,98 @@
+from functools import singledispatch
+from typing import Any, Union
+
+KEY = "__bur_serde__"
```
elijahbenizzy, line 4: burr_serde

burr/tracking/common/models.py
```diff
@@ -146,30 +136,8 @@ class BeginEntryModel(IdentifyingModel):
...
-    elif hasattr(d, "to_document"):
-        # langchain can have things that look like a document but aren't...
-        return _serialize_object(d.to_document())
-    elif hasattr(d, "model_dump"):  # generic pydantic object
-        return d.model_dump()
-    elif hasattr(d, "to_json"):
-        return d.to_json()
-    return d
+        # return _serialize_object(_filter_inputs(inputs))
+        return _filter_inputs(inputs)
```
elijahbenizzy, line 140: TODO -- move it up to the hook so we filter before serializing

elijahbenizzy (approved):
looks good, docs + small items

docs/concepts/serde.rst
```diff
@@ -0,0 +1,72 @@
...
+            myclass_kwargs = {}
+        return {
+            # required to identify how to deserialize
+            serde.KEY: "myclass",
+            # delegate to your custom serialization
+            "value": some_custom_serialization(value, **myclass_kwargs),
+        }
+
+    @serde.deserializer.register("myclass")
+    def deserialize_pickle(value: dict, myclass_kwargs: dict = None, **kwargs) -> cls:
```
elijahbenizzy, line 53: This should be called `deserialize_myclass` right?

docs/concepts/serde.rst
```diff
@@ -0,0 +1,72 @@
+================================
```
elijahbenizzy, line 1: Feels like this should be part of state persistence, we have too many concepts.

elijahbenizzy (approved):

PR #178, opened 2024-05-12
Fixes issues with streaming + sequence ID

elijahbenizzy:
https://github.com/DAGWorks-Inc/burr/issues/177

This ensures the following:

1. That increment_sequence_id is always called first
2. That it is only called once
3. That all pre run step hooks are called only once
4. That all post run step hooks are called only once

We'll likely want to improve this, the control flow is complex.

### DIFF
--- a/burr/core/application.py
+++ b/burr/core/application.py
@@ -54,7 +54,6 @@ class Transition:
 PRIOR_STEP = "__PRIOR_STEP"
 SEQUENCE_ID = "__SEQUENCE_ID"
 
-
 ERROR_MESSAGE = (
     "-------------------------------------------------------------------\n"
     "Oh no an error! Need help with Burr?\n"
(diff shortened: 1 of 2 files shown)


elijahbenizzy:
Decided to add to existing testss, as these are more "lifecycle tests" than unit tests. The logic is complicated and works on state, so there's not that much more we can do to break it up...

burr/core/application.py
```diff
@@ -867,24 +874,28 @@ def callback(
                 result: Optional[dict],
                 state: State,
                 exc: Optional[Exception] = None,
-                seq_id=self.sequence_id,
+                # seq_id=self.sequence_id,
```
skrawcz, line 877: this supposed to be commented?

skrawcz: delete otherwise?

burr/core/application.py
```diff
@@ -836,18 +848,10 @@ def streaming_response(state: State, prompt: str) -> Generator[dict, None, Tuple
             # For context, this is specifically for the case in which you want to have
             # multiple terminal points with a unified API, where some are streaming, and some are not.
             if next_action.name in halt_before and next_action.name not in halt_after:
-                self._adapter_set.call_all_lifecycle_hooks_sync(
-                    "post_run_step",
```
skrawcz, line 840: I assume this is being called somewhere else, and this was being called twice?

skrawcz (approved):
I assume it works!

burr/core/application.py
elijahbenizzy: Yep, not meant to be there (it made things more complicated)

burr/core/application.py
elijahbenizzy: Yep, we delegate to `run`

PR #180, opened 2024-05-13
Adds support for lc documents & similar

skrawcz:
LC documents can now be serialized properly.
Note: LC produces things that can look like documents that aren't. E.g. _DocumentWithState

This is a short term fix here. The longer term approach is a modular approach & structure.

## Changes
 - how LC documents are serialized, and the edge case of docs that aren't real docs...

## How I tested this
 - locally

## Checklist


elijahbenizzy (approved):
Good for a temporary fix

PR #196, opened 2024-05-22
Updates Burr tracking client to safely encode json

elijahbenizzy:
There are often missing characters -- this enables the user to safely encode them. They'll get replaced, but generally that's OK (and certainly better than failing).

Hard to test, but it doesn't introduce a regression so I'm happy enough for now.

### DIFF
--- a/burr/tracking/client.py
+++ b/burr/tracking/client.py
@@ -209,7 +209,12 @@ class LocalTrackingClient(
         **future_kwargs: Any,
     ):
         self._ensure_dir_structure(app_id)
-        self.f = open(os.path.join(self.storage_dir, app_id, self.LOG_FILENAME), "a")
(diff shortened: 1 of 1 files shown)


skrawcz (approved):
no test case?

burr/tracking/client.py
```diff
@@ -57,6 +57,12 @@ def _allowed_project_name(project_name: str, on_windows: bool) -> bool:
     return bool(re.match(pattern, project_name))
 
 
+def safe_json_write(line: str, f: io.TextIO):
```
skrawcz, line 60: where is this used?

elijahbenizzy:
> no test case?

See note. Could probably add weird bytes and serialize but its too convoluted for a small edge case IMO.

PR #222, opened 2024-06-01
Adds GraphDB Conversational Agent example

skrawcz:
With help from the FalkorDB folks we added a GraphDB example to Burr.

This uses Hamilton to ingest and push data to FalkorDB that you run locally in a docker container. Then you
can run a Burr application to run the conversational agent.

I decided to add to the conversational agent directory because that seems like the more logical place to curate conversational agent examples.

So I moved the original example to simple_example and updated links accordingly. As long as we didn't link directly to the notebook but the repository example we should be good.

## Changes
 - examples/conversational_rag

## How I tested this
- Locally

## Notes
 - Doesn't work in collab because you need to talk to the DB

## Checklist


elijahbenizzy (approved):
lgtm

PR #223, opened 2024-06-01
Updates README to use absolute links

elijahbenizzy:
Uses absolute links for the README so it can render in pypi. See feedback here: https://www.reddit.com/r/LocalLLaMA/comments/1d4p1t6/comment/l6lm56u/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

Local development will be a little annoying but we can figure it out later.


skrawcz (approved):
this wont get updated until a package is pushed.

PR #228, opened 2024-06-03
Adds field level SERDE control for state objects

skrawcz:
This allows one to globally register field level
serializers for extra customization.

The contract is that the functions need to take kwargs.

Otherwise serialize function needs to return  a dictionary.

Added tests and integration tests validating that field level is higher in priority over regular type based SERDE.

```python

def my_field_serializer(value: MyType, **kwargs) -> dict:
    serde_value = _do_something_to_serialize(value)
    return {"value": serde_value}

def my_field_deserializer(value: dict, **kwargs) -> MyType:
    serde_value = value["value"]
    return _do_something_to_deserialize(serde_value)

register_field_serde("my_field", my_field_serializer, my_field_deserializer)
```
You need to register this before running your application.

## Changes
 - state
 - tests for state

## How I tested this
 - locally

## Notes
 - Need to add documentation.

## Checklist


burr/core/state.py
```diff
@@ -196,22 +235,31 @@ def get_all(self) -> Dict[str, Any]:
...
-            for k, v in _dict.items()
-        }
+
+        def _choose_serde(k, v, **extrakwargs) -> Union[dict, str]:
+            """chooses the correct serde function for the given key"""
+            if k in FIELD_SERIALIZATION:
+                result = FIELD_SERIALIZATION[k][0](v, **extrakwargs)
+                if not isinstance(result, dict):
+                    raise ValueError(
+                        f"Field serde for {k} must return a dict, but return {type(result)}"
```
elijahbenizzy, line 245: print the actual result too

burr/core/state.py
```diff
@@ -2,13 +2,52 @@
...
+from typing import Any, Callable, Dict, Iterator, Mapping, Union
 
 from burr.core import serde
 
 logger = logging.getLogger(__name__)
 
+FIELD_SERIALIZATION = {}
+
+
+def register_field_serde(field_name: str, serializer: Callable, deserializer: Callable):
```
elijahbenizzy, line 16: We don't need nested stuff but might we want it at some point? How hard would it be to add?

burr/core/state.py
```diff
@@ -2,13 +2,52 @@
...
+    """
+    # assert that the serializer has **kwargs argument; it also needs to return a dict but we can't check that.
+    # def name(value: Any, **kwargs) -> dict
+    serializer_sig = inspect.signature(serializer)
+    if not any(param.name == "kwargs" for param in serializer_sig.parameters.values()):
+        raise ValueError(f"Serializer for [{field_name}] must have **kwargs argument.")
+
+    # assert that the deserializer has **kwargs argument; it also needs to return a dict but we can't check that.
+    deserializer_sig = inspect.signature(deserializer)
+    if not any(param.name == "kwargs" for param in deserializer_sig.parameters.values()):
```
elijahbenizzy, line 46: The better way to do this is to assert that the first parameter is of kind `VAR_KEYWORD`. Can also check its name (but I'm not sure that's relevant?).

```python
In [1]: def foo(**kwargs):
   ...:     pass
   ...:
In [2]: inspect.signature(foo).parameters['kwargs'].kind
Out[2]: <_ParameterKind.VAR_KEYWORD: 4>

In [13]: inspect.Parameter.VAR_KEYWORD
Out[13]: <_ParameterKind.VAR_KEYWORD: 4>
```

burr/core/state.py
```diff
@@ -196,22 +235,31 @@ def get_all(self) -> Dict[str, Any]:
     def serialize(self, **kwargs) -> dict:
         """Converts the state to a JSON serializable object"""
         _dict = self.get_all()
-        return {
-            # TODO: handle field specific custom serialization
-            k: serde.serialize(v, **kwargs)
-            for k, v in _dict.items()
-        }
+
+        def _choose_serde(k, v, **extrakwargs) -> Union[dict, str]:
```
elijahbenizzy, line 239: Better name than `_choose_serde`? It does more than choosing (I think)

elijahbenizzy (commented):
Looks good, a few nits

burr/core/state.py
skrawcz: People can reuse the serde module for that if they want. Otherwise not a framework concern.

skrawcz:
https://www.loom.com/share/cfee35fa59f74affb3e6114035a3e7f9?sid=b2032069-bc8b-45ac-b6b9-4866a232810b video that will upload to youtube once this is merged.

burr/core/state.py
skrawcz: Oh sure:
```python
if not any(param.kind == param.VAR_KEYWORD for param in deserializer_sig.parameters.values()):
```
checks for the right thing

burr/core/state.py
skrawcz: why? it's either None, or not a dictionary...

burr/core/state.py
skrawcz: sure.

burr/core/state.py
elijahbenizzy: It helps debugging -- I've had this exact same type of error before, and my first step is always to figure what the actual result is. Not strictly necessary, but gives the context users need.

skrawcz: sure, fine. limiting size of str.

elijahbenizzy (approved):
ship it

Issue #253, opened 2024-06-23
CRAG Example


HamzaFarhan:
**Is your feature request related to a problem? Please describe.**
Thoughts on adding **Corrective Retrieval Augmented Generation (CRAG)** as an example?

**Describe the solution you'd like**
Using `Burr`, `LanceDB`, `Instructor`, and [Exa](https://exa.ai/)(for web search). No frameworks, the SDKs directly.

**Describe alternatives you've considered**
There is a [LangGraph tutorial](https://langchain-ai.github.io/langgraph/tutorials/rag/langgraph_crag/) with `Tavily` and `Chroma` but it skips some steps and could be better overall.

**Additional context**
Here's the overview image from the [paper](https://arxiv.org/pdf/2401.15884)
<img width="1233" alt="crag" src="https://github.com/DAGWorks-Inc/burr/assets/29212086/445a9e3a-89e7-414a-87f4-191eacc4399c">

skrawcz:
@HamzaFarhan sounds good! Happy to provide input / feedback wherever you need it.

HamzaFarhan:
# Feedback Request

The possible routes that are guaranteed using `Instructor`:
1. names of all available LanceDB tables.
2. web_search
3. assistant
4. terminate (if the `query` is in ["exit", "quit", "q"])

So the `transitions` check for routes 2-4 and we assume that it must be _one_ of the tables otherwise.
`ask_assistant` will combine all `lancedb_results` and `exa_search_results` for context if they exist in the `state`, and then reset them to `[]`
We stop after `ask_assistant` to get the next `query` for `router`.
This is a combination of [Adaptive-RAG](https://arxiv.org/pdf/2403.14403) and [CRAG](https://arxiv.org/pdf/2401.15884).
## Adaptive-CRAG :test_tube: 

![statemachine](https://github.com/DAGWorks-Inc/burr/assets/29212086/41afa250-a6fd-45ce-8bb3-8e624479b2fb)

skrawcz:
@HamzaFarhan at a high level this makes sense. Not sure I have feedback without actually playing with it.  Did you have some alternate designs / thoughts?

elijahbenizzy:
This example has been merged, thank you!

PR #258, opened 2024-07-02
Adaptive CRAG example

HamzaFarhan:
Pull Request for #253


elijahbenizzy:
@HamzaFarhan thank you! On first glance, this is looking great. Will look more in-depth tomorrow then we can get it merged!

examples/adaptive_crag/application.py
```diff
@@ -0,0 +1,537 @@
...
+BURR_DOCS_DIR = Path("burr_docs")
+DOC_THRESH = 512
+LANCE_URI = "lance/rag"
+EMS_MODEL = "BAAI/bge-small-en-v1.5"  # for the embeddings in LanceDB
+DEVICE = "cuda"  # or "cpu"
+DOCS_LIMIT = 5  # number of documents to retrieve from the database or search engine
+ATTEMPTS = 3  # ask_gemini will try to call `create` this many times before giving up
+
+
+def chat_message(role: str, content: str) -> dict[str, str]:
```
elijahbenizzy, line 35: Thoughts on starting these helper functions with `_` to visually differentiate them from the Burr functions?

examples/adaptive_crag/application.py
```diff
@@ -0,0 +1,537 @@
...
+    """
+    print(f"Merging {len(docs)} docs")
+    merged_docs = [{"text": docs[0], "file_name": file_name}]
+    for doc in docs[1:]:
+        last_text = merged_docs[-1]["text"]
+        if len(last_text) <= doc_thresh:
+            merged_docs[-1]["text"] = last_text.strip() + "\n" + doc
+        else:
+            merged_docs.append({"text": doc, "file_name": file_name})
+    print(f"Merged to {len(merged_docs)} docs\n")
```
elijahbenizzy, line 66: Remove printline/replace with `logger.info`

examples/adaptive_crag/application.py
```diff
@@ -0,0 +1,537 @@
...
+
+    # define the schema of your data using Pydantic
+    class LanceDoc(LanceModel):
+        text: str = lance_model.SourceField()
+        vector: Vector(dim=lance_model.ndims()) = lance_model.VectorField()  # type: ignore
+        file_name: str
+
+    lance_db = lancedb.connect(LANCE_URI)
+
+    burr_docs = load_burr_docs(burr_docs_dir=BURR_DOCS_DIR)
```
elijahbenizzy, line 526: We should expose this separately as a function with a small CLI. Then put it in the README.

Note it needs to be done, in advance, to get it to work. Maybe add a check in the Burr workflow to see if it exists?

examples/adaptive_crag/burr_docs/actions.txt
```diff
@@ -0,0 +1,166 @@
+Actions
```
elijahbenizzy, line 1: This is fine for now, I bet we could get it to query for the docs/download them all so we don't have to rely on them being updated :) That might be better as a TODO.

examples/adaptive_crag/notebook.ipynb
elijahbenizzy, line 1: Why `time.sleep` in the while loop?

examples/adaptive_crag/application.py
```diff
@@ -0,0 +1,537 @@
...
+    except Exception as e:
+        print(f"Error in ask_assistant: {e}")
+        response = "Sorry, please try again."
+    return (
+        {"assistant_response": response},  # type: ignore
+        state.append(chat_history=user_message(query))
+        .append(
+            chat_history=assistant_message(response)  # type: ignore
+        )
+        .update(lancedb_results=[], exa_search_results=[]),
```
elijahbenizzy, line 444: Why are the lance_db/exa results written as empty? Is this to allow a reset/refresh? Worth adding a comment.

examples/adaptive_crag/application.py
```diff
@@ -0,0 +1,537 @@
...
+            + "\n<additional_context>\n"
+            + context
+            + "\n</additional_context>"
+        )
+    messages.append({"role": "user", "content": query.strip()})
+    try:
+        response = ask_gemini.create(
+            messages=messages, response_model=str, max_retries=attempts
+        )
+    except Exception as e:
```
elijahbenizzy, line 435: Which exceptions are we looking for? MIght be nice to narrow down.

elijahbenizzy (commented):
Really thorough, nice work. A few small comments -- biggest one is separating out the data ingestion piece + documenting to make it easier to run.

Otherwise I'm going to get an API key and play around to ensure it works well. Nice work!

HamzaFarhan:
Noted. Will try to update over the weekend. 👍

examples/adaptive_crag/application.py
HamzaFarhan: We get "InstructorRetryException" from Instructor after the attempts have been exhausted. Should we specify that here?

examples/adaptive_crag/application.py
HamzaFarhan: Is the @action decorator not enough of a differentiation?

examples/adaptive_crag/application.py
elijahbenizzy: Yep I think that's a good call

examples/adaptive_crag/application.py
elijahbenizzy: Normally I'd say it is but for an example I thnk it might be nice to have it as `_` to bring attention to the `@action` ones. Up to you though.

elijahbenizzy:
OK, got a chance to play around with it. Nice work! This is really slick. Would you mind doing the following?
1. Run `pre-commit` (`pip install pre-commit`, `pre-commit run --all`) -- that'll help with the tests
2. Adding instructions in the README for how to get started with gemini/exa? It's pretty easy but nice to have
3. Maybe making the jupyter notebook print out a little prettier in the interactive cell? No worries if its tricky, might just be easier to play around with.

This all looks great, let's merge shrotly!


image

examples/adaptive_crag/application.py
HamzaFarhan: How do you want this flow to look?

examples/adaptive_crag/burr_docs/actions.txt
HamzaFarhan: Yeah, I agree. If we think of it as just _any_ data for this example, then it's fine.

elijahbenizzy: 👍 yeah don't want to bog us down

elijahbenizzy:
Looks great! I'm going to go ahead and merge soon then fix up some of the testing stuff (it's unclear what's happening). If you want to add a script for data generation/checking after the fact then go for it, but I don't think it should be a blocker.

elijahbenizzy (approved):
Looks good! I'll fix up tests after the fact (and move it to `adaptive-crag` cause our linting wants dashes 😆 Then feel free to add a data loading script on top of mine if you want, but not 100% essential.

examples/adaptive_crag/application.py
```diff
@@ -0,0 +1,509 @@
...
+            max_retries=attempts,
+        )
+    except Exception as e:
+        print(f"Error in router: {e}")
+        route = "assistant"
+    return {"route": route}, state.update(query=query, route=route)  # type: ignore
+
+
+@action(reads=["query"], writes=["lancedb_query"])
+def rewrite_query_for_lancedb(
```
cmitsakis, line 279: Usually at this stage you ask for a standalone question based on the chat history (see the "conversational-rag" example). Is there a reason you don't?

elijahbenizzy: Good eyes, interesting point. The chat history is used to create a message to determine the route, but not the actual query (that comes directly from the query I think). @HamzaFarhan thoughts? Might be worth a comment or a touch-up?

cmitsakis: I mean something like this: https://github.com/DAGWorks-Inc/hamilton/blob/3763ae8d9826a1704f03d3ad25f58381ef023833/contrib/hamilton/contrib/dagworks/conversational_rag/__init__.py#L19

HamzaFarhan: Good idea.
So incorporating the chat history when rewriting the query. The CRAG paper just mentioned a `Query Rewriter`, so I went with this basic approach. I could tweak the template to include `chat_history` if available.

PR #261, opened 2024-07-06
Fixes for test case creation

skrawcz:
1. Made partition key optional.
2. This was not updated to use the latest SERDE methods.
3. It was also missing the ability to plugin custom SERDE functions.

For (3) we enable one to pass in a python file, that will then be imported, or a fully qualified module name to be imported.

## Changes
- CLI file

## How I tested this
 - locally

## Notes
Need more automated way to test this...

## Checklist

### DIFF
--- a/burr/cli/__main__.py
+++ b/burr/cli/__main__.py
@@ -289,10 +289,10 @@ def create_test_case(
     """
     if serde_module:
         if os.path.exists(serde_module):
-            print(f"Importing from file {serde_module}")
+            logger.info(f"Importing from file {serde_module}")
             import_from_file(serde_module)
         else:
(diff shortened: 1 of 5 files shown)


burr/cli/__main__.py
```diff
@@ -254,7 +282,21 @@ def create_test_case(
 
     See examples/test-case-creation/notebook.ipynb for example usage.
     See examples/test-case-creation/test_application.py for details.
+
```
elijahbenizzy, line 285: Do we have any tests for this? Should we?

elijahbenizzy (commented):
Looks good -- let's just add a bit more to docs.

elijahbenizzy (approved):

PR #265, opened 2024-07-09
Add chat_history to rewrite_query_for_lancedb in adaptive-crag

HamzaFarhan:
As commented in #258

### DIFF
--- a/examples/adaptive-crag/application.py
+++ b/examples/adaptive-crag/application.py
@@ -146,8 +146,8 @@ class LanceDBQuery(BaseModel):
         return ", ".join(self.keywords) + ", " + self.query
 
 
-def lancedb_query_template(query: str) -> str:
-    return f"""\
+def lancedb_query_template(query: str, chat_history: list[dict[str, str]]) -> str:
+    prompt = f"""\
 <task>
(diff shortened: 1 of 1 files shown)


examples/adaptive-crag/application.py
```diff
@@ -161,6 +161,13 @@ def lancedb_query_template(query: str) -> str:
 </query>
 """
 
+    if chat_history:
+        chat_history_str = "\n".join(
+            [f"{msg['role']}: {msg['content']}" for msg in chat_history]
+        ).strip()
+        prompt += f"\n<chat_history>\n{chat_history_str}\n</chat_history>"
```
elijahbenizzy, line 168: Why this format? Mind a comment?

HamzaFarhan: Anthropic's models work better with xml tags like this. And I haven't experienced any degradation with other models. So I just prefer to use them.

elijahbenizzy: Got it! Might be nice adding a comment to the code about that, but yeah, that makes sense.

elijahbenizzy (approved):

PR #286, opened 2024-07-28
Adds option to remove thread check on sqlite persister for streamlit

elijahbenizzy:
Streamlit uses multiple threads for execution, which can mess with the sqlite persister, which wants to be all on the same thread. This allows it to bypass the thread check with an option.

## Checklist

### DIFF
--- a/burr/core/persistence.py
+++ b/burr/core/persistence.py
@@ -142,16 +142,26 @@ class SQLLitePersister(BaseStatePersister):
 
     PARTITION_KEY_DEFAULT = ""
 
-    def __init__(self, db_path: str, table_name: str = "burr_state", serde_kwargs: dict = None):
+    def __init__(
+        self,
+        db_path: str,
+        table_name: str = "burr_state",
+        serde_kwargs: dict = None,
+        connect_kwargs: dict = None,
+    ):
         """Constructor
 
         :param db_path: the path the DB will be stored.
         :param table_name: the table name to store things under.
         :param serde_kwargs: kwargs for state serialization/deserialization.
+        :param connect_kwargs: kwargs to pass to the sqlite3.connect method.
+            Use check_same_thread=False to enable use ina  multithreaded context
         """
         self.db_path = db_path
         self.table_name = table_name
-        self.connection = sqlite3.connect(db_path)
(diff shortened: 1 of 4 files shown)


burr/core/persistence.py
```diff
@@ -142,16 +142,24 @@ class SQLLitePersister(BaseStatePersister):
 
     PARTITION_KEY_DEFAULT = ""
 
-    def __init__(self, db_path: str, table_name: str = "burr_state", serde_kwargs: dict = None):
+    def __init__(
+        self,
+        db_path: str,
+        table_name: str = "burr_state",
+        serde_kwargs: dict = None,
+        allow_multithreading: bool = False,
```
skrawcz, line 150: we should just pass in kwargs for connect options.

burr/core/persistence.py
```diff
@@ -142,16 +142,24 @@ class SQLLitePersister(BaseStatePersister):
...
+    ):
         """Constructor
 
         :param db_path: the path the DB will be stored.
         :param table_name: the table name to store things under.
         :param serde_kwargs: kwargs for state serialization/deserialization.
+        :param allow_multithreading: whether to allow multithreading -- this is good for streamlit, but use at your own risk
```
skrawcz, line 157: I would rename this to `connect_kwargs` and enable one to pass in everything you can to https://docs.python.org/3/library/sqlite3.html#sqlite3.connect

skrawcz (changes requested):
Just go ahead and expose passing in connect kwargs. That will simplify things. E.g. for disabling the check for the same thread see kwargs you can pass to connect...

burr/core/persistence.py
elijahbenizzy: Meh I could go either way, but fine with that

elijahbenizzy: At that point we might as well just pass in the connection...

skrawcz: > At that point we might as well just pass in the connection...

You could create classmethods for convenience, and then have the constructor take the connection -- but that seems like a bigger change...

skrawcz (approved):

PR #291, opened 2024-07-30
Adds endpoints to debug what has been persisted

skrawcz:
This I think will be useful for production debugging; what's in my state?

This assumes:

1. You have the right dependencies with the server to load the right persister.
2. Not all persisters load via the constructor, so you can use a classmethod to be called to instantiate it.
3. Haven't exposed writing to state, but it's something we could do.

We'd probably want to couple this with a UI view.

## Changes
 - server run.py with two endpoints
 - adds pydantic models returned

## How I tested this
 - not extensively, but it seems to work.

## Checklist

### DIFF
--- a/burr/tracking/server/run.py
+++ b/burr/tracking/server/run.py
@@ -1,7 +1,7 @@
 import importlib
 import os
 from importlib.resources import files
-from typing import Sequence
+from typing import Optional, Sequence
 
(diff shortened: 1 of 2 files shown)


andreahlert:
Closing due to inactivity (20 months). The feature idea is still valid but this PR likely needs a rewrite against the current codebase. Feel free to reopen.

PR #295, opened 2024-07-31
WIP: prototyping proxy object spans

skrawcz:
So this prototypes something to connect things to the UI.

Screen Shot 2024-07-30 at 11 51 26 PM

Things to discuss creating the spans:

1. when do we proxy the object? before and after each step? (can't do it with step because post doesn't get inputs... but could?)
2. do we do it once at the beginning (would need to expose inputs)? But then how do we provide the tracer? with before and after hooks?

Things for the spans:

1. need to log what can be pulled from the client -- where? how? e.g. I want to emit tokens used.
2. how does this then display in the UI?

Otherwise technically this should work for any "object". We'd want to internally __burr_ the variables.

Lots to chat about.

## Checklist

### DIFF
--- a/burr/core/application.py
+++ b/burr/core/application.py
@@ -525,6 +525,7 @@ class Application:
                     sequence_id=self.sequence_id,
                     app_id=self._uid,
                     partition_key=self._partition_key,
+                    _ProxyClassHook__tracer=self.dependency_factory["__tracer"],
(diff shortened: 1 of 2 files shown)


elijahbenizzy:
Closed in favor of #316 -- that does this + more

PR #315, opened 2024-08-09
improvement: handle `None` in `list_app_ids()`

zilto:
`list_app_ids()` was the only method that expected you to magically know that the default `partition_key` is `""`. Now, it accepts `None` as an argument.

Also, took the opportunity to replace the use of the instance attribute by the class attribute.


elijahbenizzy:
Nice! This makes sense. Might be nice to have a test?  I think e have other tests for this.

elijahbenizzy (approved):
Looks good!

PR #326, opened 2024-08-16
@trace decorator for non-burr functions

elijahbenizzy:
Trace decorator for non-burr functions

## Changes

Adds `@trace` decorator to API

## How I tested this

Tests + manual

## Notes

Two things we'll want to fix:
- [ ] We instantiate the tracerfactory object every step. Low overhead (just a dataclass), but might be worth optimizing away anyway.
- [ ] We don't currently pin errors to spans, we'll want to

## Checklist

### DIFF
--- a/burr/visibility/tracing.py
+++ b/burr/visibility/tracing.py
@@ -1,10 +1,16 @@
 import dataclasses
+import functools
+import inspect
+import logging
 import uuid
 from contextlib import AbstractAsyncContextManager, AbstractContextManager
 from contextvars import ContextVar
-from typing import Any, List, Optional, Type
+from hashlib import sha256
+from typing import Any, Callable, Dict, List, Optional, Type, TypeVar
 
 from burr.lifecycle.internal import LifecycleAdapterSet
+
+logger = logging.getLogger(__name__)
 
 
 @dataclasses.dataclass
@@ -281,7 +287,7 @@ class TracerFactory(ActionSpanTracer):
         action: str,
         sequence_id: int,
         app_id: str,
-        partition_key: str,
+        partition_key: Optional[str],
         lifecycle_adapters: LifecycleAdapterSet,
         _context_var: ContextVar[Optional[ActionSpan]] = execution_context_var,
     ):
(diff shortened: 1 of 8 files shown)


elijahbenizzy:
TODO for release readiness:
- [ ] Example + add to example
- [ ] Documentation
- [ ] Clean up/polish code
- [ ] Add function-details as attributes to trace (module, code hash, etc...)
- [ ] Tests
    - [ ]  Add to our tracing tests
    - [ ] Test E2E async
    - [ ] Test E2E sync
- [ ]  Wire through failures
- [ ] Ensure serde works with the attributes
- [ ] Remove the context manager in application, should just be `.set`, `.reset()`
- [ ] Remove any other cruft

burr/visibility/tracing.py
```diff
@@ -321,3 +327,142 @@ def __call__(
...
+class trace:
+    def __init__(
+        self,
+        capture_inputs: bool = True,
+        capture_outputs: bool = True,
+        input_filterlist: Optional[List[str]] = None,
+        span_name: Optional[str] = None,
+        _context_var: ContextVar[Optional[TracerFactory]] = tracer_factory_context_var,
+    ):
+        """trace() surrounds a burr function and uses the tracer to create a span and log
```
skrawcz, line 354:
```suggestion
        """trace() can surround any function and uses the tracer to create a span and log
```

burr/visibility/tracing.py
```diff
@@ -321,3 +327,142 @@ def __call__(
...
+
+            from burr.visibility import trace
+
+            @trace()
+            def call_llm(prompt):
+                return _query(...)
+
+            @trace()
+            def generate_text(prompt: str) -> str:
+                result = call_openai(prompt)
```
skrawcz, line 373:
```suggestion
                result = call_llm(prompt)
```

burr/visibility/tracing.py
```diff
@@ -321,3 +327,142 @@ def __call__(
...
+                return f"<p>{result}</p>"
+
+            @action(reads=["prompt"], writes=["response"])
+            def prompt_action(state: State) -> State:
+                response = generate_text(state["prompt"])
+                return state.update(response=response)
+
+        Every time `prompt_action` is called, it will generate a trace that looks like the following:
+        If it is called outside the context of a Burr action, it will be effectively a no-op.
```
skrawcz, line 382:
```suggestion
        Every time `prompt_action` is called, it will generate a trace that looks like the following:
         
             prompt_action
                  generate_text
                        call_llm
        
        If it is called outside the context of a Burr action, it will be effectively a no-op.
```

examples/email-assistant/statemachine.png
skrawcz, line 1: why the extra detail on the image? Seems overwhelming to me.

skrawcz:
So people don't need to set anything when building an application right? They can just import trace and it'll just work?

elijahbenizzy:
> So people don't need to set anything when building an application right? They can just import trace and it'll just work?

As long as they have a tracker

docs/concepts/additional-visibility.rst
```diff
@@ -138,6 +138,51 @@ Note that we are currently building out the capability to wrap a class and "auto
...
+            with __tracer('check_prompt_safety'):
+                if _is_unsafe(modified_prompt):
+                    modified_prompt = _fix(modified_prompt)
+        with __tracer('call_llm', dependencies=['create_prompt']):
+            response = _query(prompt=modified_prompt)
+        return state.update({'response': response})
+
+
+This will create spans for the ``_modify_prompt``, ``_fix``, and ``_query`` functions, and log the parameters and return values.
+You can opt out of logging paramers or return values and adding a filter to the decorator to exclude certain parameters.
```
skrawcz, line 184:
```suggestion
.. code-block:: python

    from burr.visibility import trace

    @trace
    def _modify_prompt(prompt: str) -> str:
        return ...
    @trace
    def _fix(prompt: str) -> str:
        return ...

    @trace
    def _query(prompt: str) -> Response:
        return ...
   
    from burr.core import action

    @action(reads=['prompt'], writes=['response'])
    def my_action(state: State, __tracer: TracingFactory) -> State:
        modified_prompt = _modify_prompt(state["prompt"])
        if _is_unsafe(modified_prompt):
              modified_prompt = _fix(modified_prompt)
        response = _query(prompt=modified_prompt)
        return state.update({'response': response})

This will create spans for the ``_modify_prompt``, ``_fix``, and ``_query`` functions, and log the parameters and return values. You can opt out of logging paramers or return values and adding a filter to the decorator to exclude certain parameters.

To contrast what you can instrument manually, the `@trace` decorator allows you to get more visibility and replace manual code such as the following (though you can happily combine the two approaches):

.. code-block:: python

    from burr.visibility import TracingFactory
    from burr.core import action

    @action(reads=['prompt'], writes=['response'])
    def my_action(state: State, __tracer: TracingFactory) -> State:
        with __tracer('create_prompt'):
            modified_prompt = _modify_prompt(state["prompt"])
            with __tracer('check_prompt_safety'):
                if _is_unsafe(modified_prompt):
                    modified_prompt = _fix(modified_prompt)
        with __tracer('call_llm', dependencies=['create_prompt']):
            response = _query(prompt=modified_prompt)
        return state.update({'response': response})
```

The `with __tracer` stuff is overkill. So clarifying the example here to be simpler.

skrawcz (approved):

PR #330, opened 2024-08-17
scratch pad for fixing partition key with local tracker

elijahbenizzy:


elijahbenizzy:
This ensures we live in two worlds. Other options:
1. Auto-migrate on opening
2. Add it to app_id
3. TBD (?)


TODO if we want this version:
- [ ] Get it working in the server (shared code maybe?)
- [ ] Add testing/test manually
- [ ] Document out what happens

andreahlert:
Closing. Same scope as #519. The partition key fix landed via PR #682.

PR #343, opened 2024-08-28
Fixes up the s3 server

elijahbenizzy:
Errors when we had too many + errors around ordering + pagination

### DIFF
--- a/burr/tracking/server/s3/backend.py
+++ b/burr/tracking/server/s3/backend.py
@@ -317,7 +317,10 @@ class SQLiteS3Backend(BackendBase, IndexingBackendMixin, SnapshottingBackendMixi
         return out
 
     async def _gather_paths_to_update(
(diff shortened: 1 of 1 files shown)


burr/tracking/server/s3/backend.py
```diff
@@ -340,8 +343,10 @@ async def _gather_paths_to_update(
                     # Created == last_modified as we have an immutable data model
                     logger.info(f"Found new file: {key}")
                     paths_to_update.append(DataFile.from_path(key, created_date=last_modified))
+                    if len(paths_to_update) >= max_paths:
+                        break
         logger.info(f"Found {len(paths_to_update)} new files to index")
-        return paths_to_update
+        return paths_to_update[:max_paths]
```
skrawcz, line 349: why is this needed -- the break above should suffice?

skrawcz (approved):

PR #344, opened 2024-08-28
Adds additional conditions/transitions

elijahbenizzy:
condition.lmda is just a shortcut for Condition. This also has a commented out condition.exists -- this is ready, but we do not support optional state items, so it is not feasible yet.

## Checklist


skrawcz (approved):
Add to docs somewhere?

elijahbenizzy:
> Add to docs somewhere?

Waiting until docs change up Thierry is doing

Issue #342, opened 2024-08-28
Pagination in the UI


elijahbenizzy:
**Is your feature request related to a problem? Please describe.**
It's just on the server side

**Describe the solution you'd like**
We want it on the UI side as well

**Describe alternatives you've considered**
Should be simple

PR #346, opened 2024-08-28
Adds pagination to apps view page

elijahbenizzy:
This is not perfect -- but it'll do. The caveat is that parent/child pointers aren't always found -- we need to add that to the API.

## Checklist

### DIFF
--- a/burr/tracking/server/backend.py
+++ b/burr/tracking/server/backend.py
@@ -4,7 +4,7 @@ import importlib
 import json
 import os.path
 import sys
-from typing import Any, Optional, Sequence, Type, TypeVar
+from typing import Any, Optional, Sequence, Tuple, Type, TypeVar
 
 import aiofiles
 import aiofiles.os as aiofilesos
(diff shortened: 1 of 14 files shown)


skrawcz:
@elijahbenizzy node.js error

elijahbenizzy:
#342

skrawcz:
> The caveat is that parent/child pointers aren't always found -- we need to add that to the API.

New issue?

elijahbenizzy:
image

burr/tracking/server/backend.py
```diff
@@ -230,8 +235,13 @@ async def _load_metadata(self, metadata_path: str) -> models.ApplicationMetadata
...
     async def list_apps(
-        self, request: fastapi.Request, project_id: str, partition_key: Optional[str]
-    ) -> Sequence[ApplicationSummary]:
+        self,
+        request: fastapi.Request,
+        project_id: str,
+        partition_key: Optional[str],
+        limit: int = 100,
+        offset: int = 0,
+    ) -> Tuple[Sequence[ApplicationSummary], bool]:
```
skrawcz, line 244: bool? this should be int right?

burr/tracking/server/backend.py
```diff
@@ -101,8 +101,13 @@ async def list_projects(self, request: fastapi.Request) -> Sequence[schema.Proje
...
     async def list_apps(
-        self, request: fastapi.Request, project_id: str, partition_key: Optional[str]
-    ) -> Sequence[schema.ApplicationSummary]:
+        self,
+        request: fastapi.Request,
+        project_id: str,
+        partition_key: Optional[str],
+        limit: int,
+        offset: int,
+    ) -> Tuple[Sequence[schema.ApplicationSummary], int]:
```
skrawcz, line 110: adjust doc string to say why int

skrawcz:
> image

where is the pagination?

elijahbenizzy:
image

One left todo -- add ... to show that we have more pages. In the code there's aline on where to do it.

skrawcz (approved):

PR #348, opened 2024-08-29
`docs` Created a `Cookbook` section

zilto:
Creating the cookbook section involved:

- refactoring the docs navigation bar
- reordering examples and grouping them
- copy-pasting `README` and notebooks from `/examples` into `/docs` (because sphinx can't see documents outside of its root)
- add the sphinx extension `myst_nb` to parse notebooks (it's redundant with myst_parser, so I removed it)
- Added narrative content to some of the notebooks that didn't have much context (by copying text from the README). This could be improved further for some cookbook entries (e.g., `cowsay`)

Note
- when using `.ipynb` format, the header 1 level (`#`) should be reserved to the title of the notebook in the first cell, then subsequent headers should start at level 2 (`##`)


elijahbenizzy:
Overall thoughts -- not all reltaed to these changes, but changes that are worth making:
- [ ] We should make "why burr" more succinct. Shouldn't mention state machines. Instead "Burr gives you all of these for free, keep reading for more".
- [ ] we should update images to have the new graph format, or UI screencaps
- [ ] Ensure all `pip install burr[X]` has quotes -- `pip install "burr[X]"`
- [ ] Should have a "how to learn" in "next steps" -- walking through docs/recommending what to do
- [ ] We shoudl remove streaming results from the "planned capabilities" -- it is implemented!
- [ ] Two more cookbook examples we'll need
    - [ ] streaming
    - [ ] human in the loop
- [ ] Similarly, we need a way to represent more concrete concepts/how-tos, Thinking:
    - [ ] We should move most of the concept stuff to references
    - [ ] We should add concepts as a "journey" -- E.G. define an action, add human input, wire together in an application, add streaming output, expose in a web-service, etc...

skrawcz (approved):
As always -- lots of things could be added and improved. I think this is sufficient to merge.
Only question is around adding google colab link in the notebooks.

zilto:
> Only question is around adding google colab link in the notebooks.

Agreed, but I didn't want to make significant content changes in the copied examples in this PR. Adding Colab links would require adding the `!pip install` command in all notebooks.

PR #349, opened 2024-08-31
Test case tweak

skrawcz:
Clarifies test case creation behavior and adds ability to override the action name

## Changes
 - CLI command
 - docs
 - fixes some links

## How I tested this
 - locally 

## Checklist


burr/cli/__main__.py
```diff
@@ -134,8 +134,10 @@ def _run_server(
     cmd = f"uvicorn burr.tracking.server.run:app --port {port} --host {host}"
     if dev_mode:
         cmd += " --reload"
-    base_dir = os.path.expanduser("~/.burr")
-    if not no_copy_demo_data:
+
+    if backend == "local" and not no_copy_demo_data:
```
elijahbenizzy, line 138: Best to instantiate the settings module if we can -- that way it's more future-proof/solves your TODO

burr/cli/__main__.py
```diff
@@ -322,13 +333,19 @@ def import_from_file(file_path: str) -> ModuleType:
     required=False,
     help="Python file or fully qualified python module to import for custom serialization/deserialization.",
 )
+@click.option(
```
elijahbenizzy, line 336: Why do we need to change this? A little unclear what this does -- worth adding a bit more docs

burr/cli/__main__.py
```diff
@@ -358,17 +383,60 @@ def create_test_case(
...
+    Explanation of the logic here.
+    If there's an error:
+     - status would be failed vs completed
+     - state loaded would be the starting state
+    If there's no error:
+     - status would be completed
+     - state loaded would be the ending state, so to get the starting state, we need to look at the prior sequence ID
+    """
+    if after_action["status"] == "completed":
+        print("Action was successful so loading initial and expected state into test fixture.")
```
elijahbenizzy, line 406: We should have logger.warn instead of print here as we use stdout.

burr/tracking/server/backend.py
```diff
@@ -189,9 +189,10 @@ def get_uri(project_id: str) -> str:
 
 
 class LocalBackend(BackendBase):
-    """Quick implementation of a local backend for testing purposes. This is not a production backend."""
+    """Quick implementation of a local backend for testing purposes. This is not a production backend.
 
-    # TODO -- make this configurable through an env variable
+    To override the path, set a `burr_path` environment variable to the path you want to use.
```
elijahbenizzy, line 194: Point out the settings class that corresponds to this

elijahbenizzy (commented):
Looks good, some nits

burr/cli/__main__.py
skrawcz: it provides the name of the action to test for the test case.

enables you to use output from somewhere else as input for another action easily.

burr/cli/__main__.py
skrawcz: prints are used throughout here.

elijahbenizzy: We shouldn't -- it sometimes will/sometimes won't parse, and you'll be putting that to a file. You want to be able to pipe it.

elijahbenizzy (approved):

PR #355, opened 2024-09-06
Prompt playground

zilto:
Allows to load traces from tracked Burr application with OpenLLMetry traces.

You can load previous chat interactions and try new prompts with multiple LLM providers via LiteLLM. 

image

To launch the app

`OPENAI_API_KEY=sk-... ANTHROPIC_API_KEY=sk-ant-... streamlit run burr/integrations/playground/app.py`

Features:
- 3 providers at once
- load interactions directly from local Burr tracking
- load a specific interaction via the URL params
- the playground interactions themselves are logged using Burr under `burr-playground`

Limitations:
- streamlit makes it very inconvenient to call LLMs asynchronously
- currently no streamlined way to support generic API key input

### DIFF
--- a/burr/integrations/playground/app.py
+++ b/burr/integrations/playground/app.py
@@ -188,7 +188,8 @@ def history_component(normalized_spans):
         answer = normalized_spans["gen_ai"]["completion"].values()
 
         for message in list(prompts) + list(answer):
-            with st.chat_message(message["role"]):
+            role = message.get("role", "assistant")
+            with st.chat_message(role):
                 st.markdown(message["content"])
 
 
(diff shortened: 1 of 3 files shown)


elijahbenizzy:
OK, looking good. A few minor UI points after playing around:
1. Model not selected doesn't run it -- maybe we should be able to schose an arbitrary number of them? Or just grey it out if it's not selected?
2. Launch button is ina. weird place, I think it should be below the provider/model selection
3. Selecting a model resets everything -- it shouldn't reset the other ones -- we shoud have a reset button
4. We should have a progress button if it's running
5. The third one never seems to work for me...
6. Tab to link to the burr UI (iframe?)
See what I'm working with:

image

elijahbenizzy (commented):
Some thoughts here -- a few nits on the structure then feedback on the UI

Issue #359, opened 2024-09-08
1-to-many transitions


zilto:
# Currently
Below is a valid Burr graph definition to "get a user input, and select the most appropriate action out of 3"

```python
graph = (
  GraphBuilder()
  .with_actions(
    process_user_input,
    decide_next_action,
    generate_text,
    generate_image,
    ask_for_details,
  )
  .with_transitions(
      ("process_user_input", "decide_next_action"),
      ("decide_next_action", "generate_text"),
      ("decide_next_action", "generate_image", expr("mode==generate_image")),
      ("decide_next_action", "ask_for_details", expr("mode==ask_user_for_details")),
  )
  .build()
)
```
Notes:
- Needs a "decide node" that writes a value to state (i.e., `mode`). 
- The "decide node" is the result of `Condition` objects requiring to return `bool`
- Has one `transition` per graph edge
- Requires at least 2 out of 3 explicit `condition`
- Conditions are evaluated sequentially 

# Desired solution
The main benefit of the above is that everything is explicit. But it can also be less ergonomic, add complexity to defining transitions, and be inefficient when computing transitions is expensive. 

Consider the following API
```python
ApplicationBuilder()
  .with_actions(
    process_user_input,
    generate_text,
    generate_image,
    ask_user_for_details,
  )
  .with_transitions(
    ("process_user_input", "decide_next_action"),
    (
      "decide_next_action",
      ["generate_text", "generate_image", "ask_user_for_details"],
      OneToManyCondition(...)   # TODO
    ),
  )
  .with_entrypoint("process_user_input")
  .build()
)
```

Note:
- No longer requires a "decide node"; this is moved to a `OneToManyCondition` object. When "resolved", it should return the index or the name of the next action.
- New `transition` syntax allowing `1 -> n`. This allows to resolve multiple binary conditions at once
- arguably easier to read and manage code changes
- less transitions to define

# Use case
The popular use case I have in mind is "use an LLM to decide the next node". Given the `Graph` building process, it would be possible to dynamically create a model of "available next actions". Here's a sketch using `instructor`, which has better guarantees regarding structured LLM outputs (for OpenAI at least)

```python
ApplicationBuilder()
  .with_actions(
    process_user_input,
    generate_text,
    generate_image,
    ask_user_for_details,
  )
  .with_transitions(
    ("process_user_input", "decide_next_action"),
    (
      "decide_next_action",
      ["generate_text", "generate_image", "ask_user_for_details"],
[...]

PR #361, opened 2024-09-09
added `Select` to define transitions

zilto:
Following #359, this PR introduces `Select` as an alternative to `Condition` when defining "one to many" transitions.

Not only can this be more ergonomic for developers, it allows to resolve multiple conditions at once. This was impossible before and conditions were only binary checks evaluated sequentially. 

For example, this is now a valid `Graph` definition

```python
import random
from burr.core import action, ApplicationBuilder, State, Select, Action

def random_decider(state: State, actions: list[Action]) -> str:
    action_idx = random.randint(0, len(actions)-1)
    return actions[action_idx].name

app = (
    ApplicationBuilder()
    .with_actions(
        process_user_input, generate_text, generate_image,
        ask_user_for_details, send_response
    )
    .with_transitions(
        (
            "process_user_input",
            ("generate_text", "generate_image", "ask_user_for_details"),
            Select(["user_query"], resolver=random_decider),
        ),
        (
            ("generate_text", "generate_image", "ask_user_for_details"),
            "send_response"
        ),
        ("send_response", "process_user_input"),
    )
    .with_entrypoint("process_user_input")
    .build()
)
```
decide

## Changes
- Added the `Select` construct
- Updated the `.get_next_node()` and `.with_transitions()` of `Graph` and `GraphBuilder` to support `Select`
- Updated the `GraphBuilder()` and `ApplicationBuilder()` type annotations to receive `Select`
- Updated the `GraphBuilder()` and `ApplicationBuilder()` type annotations to use `Sequence` instead of `List`

## How I tested this
- Added two tests to create `Select` objects and call the `.run()` method
- all previous tests are still passing

## Potential TODOs
- add a `default` kwarg to `Select`. This could be the default `Action` to use if the `resolver` returns `None` for instance
- Allow the `resolver` to directly return an `Action` or an integer index of the action instead of `action_name: str`
- Properly type the `Select.resolver` because it currently dislikes user-provided functions annotated with `list` instead of `Sequence`

## TODOs
- Add `Select` to the main docs
- Add docstring documentation with examples (where exactly? directly on `Select`? Should I update `.with_transitions()` of other objects?)
- Create an `examples/` that uses an LLM to decide the next action (see #359) 

## Checklist

### DIFF
--- a/burr/core/__init__.py
+++ b/burr/core/__init__.py
@@ -1,4 +1,4 @@
-from burr.core.action import Action, Condition, Result, action, default, expr, when
+from burr.core.action import Action, Condition, Result, Select, action, default, expr, when
 from burr.core.application import (
     Application,
     ApplicationBuilder,
@@ -18,6 +18,7 @@ __all__ = [
     "default",
     "expr",
     "Result",
+    "Select",
     "State",
     "when",
 ]
--- a/burr/core/action.py
+++ b/burr/core/action.py
@@ -17,6 +17,7 @@ from typing import (
     List,
     Optional,
     Protocol,
+    Sequence,
     Tuple,
     TypeVar,
     Union,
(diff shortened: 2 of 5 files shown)


skrawcz:
@zilto I think I could come around to this if the `select` actually became a node in the graph. My main concern is the select is logic that functions over state and would be crucial for debugging and observing. We don't do this for "lambdas" that one could write for a single edge, and here you've just collapsed it to a single function. So maybe we need a third entity type for this to be represented in the graph...

elijahbenizzy:
> @zilto I think I could come around to this if the `select` actually became a "node" (maybe a different type) in the graph. My main concern is the select is logic that functions over state and would be crucial for debugging and observing. We don't do this for "lambdas" that one could write for a single edge, and here you've just collapsed it to a single function (for each edge you could write a lambda for it that functions similarly). So maybe we need a third entity type for this to be represented in the graph...

Yeah agreed on difficulty for observing -- that said, I think it's better to just fix observability for edges. This would solve it: https://github.com/DAGWorks-Inc/burr/issues/437

PR #366, opened 2024-09-18
Copies pydantic model config in subset model for pydantic integration

elijahbenizzy:
Beforehand we were dropping this. Thus features like allowing arbitrary fields was disallowed.

## Changes
- Adds config copy

## How I tested this
- Manually 
- Unit tests

## Notes
- Thanks @zilto for the flag

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ tests = [
   "langchain_community",
   "pandas",
   "psycopg2",
-  "pydantic",
+  "pydantic[email]",
   "pyarrow",
   "redis",
   "burr[opentelemetry]"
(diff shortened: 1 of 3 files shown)


zilto (approved):
LGTM. Main failure point was `subset_model()`

pyproject.toml
```diff
@@ -58,7 +58,7 @@ tests = [
   "langchain_community",
   "pandas",
   "psycopg2",
-  "pydantic",
+  "pydantic[email]",
```
skrawcz, line 61: email?

elijahbenizzy: test

zilto: Elijah said he added it only for testing purposes :welp:

PR #370, opened 2024-09-22
Parallelism

elijahbenizzy:

## Checklist

We are going to release this in beta mode. Remaining tasks, will turn into issues
- [ ] Decide how to "label" the tasks
- [ ] Add guard-rails for async to ensure people don't mess stuff up with the async generator


docs/concepts/parallelism.rst
```diff
@@ -0,0 +1,563 @@
+===========
+Parallelism
+===========
+
+Burr allows for sets of actions/subgraphs to run in parallel. In this section we will go over the use-cases/how to run them!
```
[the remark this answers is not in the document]
elijahbenizzy: Nice, thanks

skrawcz:
question - does this support async paths?

elijahbenizzy:
> question - does this support async paths?

Yes see the points on async executor. Not fleshed out in this doc but it will be included in the first release

docs/concepts/parallelism.rst
```diff
@@ -0,0 +1,583 @@
...
+
+    class TestMultipleModels(MapActions):
+
+        def actions(self, state: State) -> Generator[Action | Callable | RunnableGraph, None, None]:
+            # make sure to add a name to the action
+            # This is not necessary for subgraphs, as actions will already have names
+            for action in [
+                query_llm.bind(model="gpt-4").with_name("gpt_4_answer"),
+                query_llm.bind(model="o1").with_name("o1_answer"),
+                query_llm.bind(model="claude").with_name("claude_answer"),
```
yiweig, line 159: The examples here use the same action with different parameters bound to it. Is it possible to instead map over different actions completely? For example something like this:

```python
@action(reads["query"], writes=["llm_output"])
def query_llm(state: State) -> State:
    return state.update(llm_output=some_function(state["query"]))

@action(reads["query"], writes=["llm_output"])
def count_tokens(state: State) -> State:
    return state.update(llm_output=another_function(state["query"]))

@action(reads["query"], writes=["llm_output"])
def draw_chart(state: State) -> State:
    return state.update(llm_output=a_third_function(state["query"]))

def actions(self, state: State) -> Generator[Action | Callable | RunnableGraph, None, None]:
    for action in [
        query_llm,
        count_tokens,
        draw_chart,
    ]:
        yield action
```

As a follow up question, is there a requirement that all the actions have to write the same keys? My naive intuition is that this is not a requirement, but I would have to do some extra work in the `reduce` function to make sure I get all available outputs:

```python
@action(reads["query"], writes=["llm_output"])
def query_llm(state: State) -> State:
    ..
@action(reads["query"], writes=["token_count"])
def count_tokens(state: State) -> State:
    ..
@action(reads["query"], writes=["chart_bytes"])
def draw_chart(state: State) -> State:
    ..

def reduce(self, states: Generator[State, None, None]) -> State:
    outputs = {}
    for state in states:
        if "llm_output" in state:
            outputs["llm_output"] = state["llm_output"]
        if "token_count" in state:
            outputs["token_count"] = state["token_count"]
        if "chart_bytes" in state:
            outputs["chart_bytes"] = state["chart_bytes"]
        return state.update(outputs)

def writes() -> List[str]:
    return ["llm_output", "token_count", "chart_bytes"]
```

Alternatively, perhaps it would be better to do this to avoid burr complaining that a particular state key was not present in the output
```python
def reduce(self, ..):
    outputs = {}
    ..
    return state.update({"outputs": outputs})

def writes():
    return ["outputs"]
```

Sorry for the long comment, just trying to wrap my head around the possibilities with the new API!

docs/concepts/parallelism.rst
```diff
@@ -0,0 +1,583 @@
...
+            ]
+                yield action
+
+        def state(self, state: State) -> State::
```
yiweig, line 163: nit: I think there's a colon `:` missing from the `for` and an extra colon at the end of the `def state(..)` :)

yiweig (commented):
First off, thanks so much for burr! We've been using it to build some agentic workflows at my job and it's been a breeze to set up and iterate.

I was directed to this PR from https://github.com/DAGWorks-Inc/burr/discussions/339, and briefly looked over the new API. I love the direction you guys are going with this!

Had a few questions related to mapping the same state over different actions because that's a use case we're interested in, let me know if any of my questions don't make sense or if you need me to clarify!

docs/concepts/parallelism.rst
elijahbenizzy: Hey! Yes, I think that's supported (If I understand correctly). It's a little nuanced (and I could explain it better), but the idea is that the generator can yield an `Action` object, a `Callable` (function that will get turned into an action, or a `RunnableGraph` (a subgraph that functions as an action). Specifically, this allows you to do exactly what you want/mix + match.

The way to reconcile this with the example (with multiple `.bind()`) calls is that the `.bind` creates a new action, so for the sake of the `actions` generator it doesn't really matter whether it was created through `.bind(...)` or created through specifying another action entirely!

Regarding the follow-up, currently, yes, it's necessary to apply defaults. This is not a feature of parallelism, rather a feature of Burr. Specifically, we want some strict sense of what an action writes so we have guarentees on what should show up next. The general approach is to apply defaults -- E.G. say if state doesn't contain `foo`, it'll be `None` -- I think the simplest in your case would be to do something like:

```python
def reduce(self, states: Generator[State, None, None]) -> State:
    outputs = {'llm_output' : None, 'token_count' : None, 'chart_bytes' : None}  # initialize so you can overwrite with results
    # This is generic, or just have it like you did above, easier to read maybe
    for state in states:
        for key in list(outputs):
            if key in state:
                output[key] = state[key]
        return state.update(outputs)
```

docs/concepts/parallelism.rst
elijahbenizzy: Thanks! Good find. Will update 🙏

docs/concepts/parallelism.rst
```diff
@@ -0,0 +1,583 @@
...
+- We pass an instance of the ``TestMultipleModels`` class to the ``ApplicationBuilder``, which will run the action over the states we provide.
+
+.. code-block:: python
+
+    from burr.core import action, state
+    from burr.core.parallelism import MapActions, RunnableGraph
+    from typing import Callable, Generator, List
+
+    @action(reads=["prompt", "model"], writes=["llm_output"])
+    def query_llm(state: State, model: str) -> State:
```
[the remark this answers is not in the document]
elijahbenizzy: Hmm, good point re: consistency! I think it's better if everything else is called `query_llm` -- this is the only one with `action` appended, and the docs switch between. Changed!

docs/concepts/parallelism.rst
yiweig: Amazing, thank you! This new API is going to be very helpful for us :)

elijahbenizzy: Great! Would love to hear about how you'd be using it/specifics (only if you feel comfortable sharing). This can help us get a sense of ergonomics (and we can send you test versions if you're interested in messing with it). Feel free to join the discord and DM us if that would be useful!

skrawcz (approved):
let's get v1 out

docs/concepts/parallelism.rst
```diff
@@ -0,0 +1,587 @@
...
+
+Execution
+---------
+
+To enable execution, you need to pass a ``burr.Executor`` to the application, or to the actions themselves. We have a few available executors:
+
+- ``burr.parallelism.MultiThreadedExecutor`` -- runs the tasks in parallel using threads (default)
+- ``burr.parallelism.MultiProcessExecutor`` -- runs the tasks in parallel using processes
+- ``burr.parallelism.RayExecutor`` -- runs the tasks in parallel using `Ray <https://docs.ray.io/en/latest/index.html>`_
+- ``burr.parallelism.Dask`` -- runs the tasks in parallel using `Dask <https://dask.org/>`_
```
skrawcz, line 454: this section should be updated to reflect what we've shipped though?

PR #412, opened 2024-10-30
fix: support `.warm_up()` in `HaystackAction`

zilto:
Fixes two important bug in `HaystackAction` 

1. The `HaystackAction` and `Application` needed a way to [warm up Haystack components](https://docs.haystack.deepset.ai/docs/components#warming-up-components). Warming up typically involves loading a model in memory. The current solution is to allow `HaystackAction(do_warm_up: bool)` to specify if the component should be warmed up at action instantiation. The value can be set to false if the `Component` was warmed up before creating the `HaystackAction`

2. The `HaystackAction.optional_and_required_inputs()` was returning a tuple of lists instead of sets. This was breaking the method `Graph.visualize()` when both optional and required inputs were present because it expects sets and uses the merge operator `|` between them.

Additional tests were added to support these fixes.

----

> [!IMPORTANT]
> Add support for component warm-up and fix input handling in `HaystackAction`.
> 
>   - **Behavior**:
>     - Add `do_warm_up` parameter to `HaystackAction.__init__()` to optionally call `Component.warm_up()`.
>     - Fix `HaystackAction.optional_and_required_inputs()` to return sets instead of lists.
>   - **Tests**:
>     - Add tests for `do_warm_up` behavior in `test_burr_haystack.py`.
>     - Add tests for `optional_and_required_inputs()` return type in `test_burr_haystack.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=DAGWorks-Inc%2Fburr&utm_source=github&utm_medium=referral)<sup> for 25338a408b50e60313ae0e42b5850733328ed481. It will automatically update as commits are pushed.</sup>


burr/integrations/haystack.py
```diff
@@ -39,8 +40,10 @@ def __init__(
         Use a mapping {state_field: socket} to rename Haystack output sockets (see example).
         :param name: Name of the action. Can be set later via ``.with_name()``
         or in ``ApplicationBuilder.with_actions()``.
-        :param bound_params: Parameters to bind to the `Component.run()` method.
-
+        :param bound_params: Parameters to bind to the ``Component.run()`` method.
+        :param is_warm: If True, try to call ``Component.warm_up()`` if it exists.
```
elijahbenizzy, line 44: This doesn't match the params? Think you just want to change the param name in the docstring.

burr/integrations/haystack.py
```diff
@@ -346,6 +357,7 @@ def haystack_pipeline_to_burr_graph(pipeline: Pipeline) -> Graph:
             socket_name: socket_mapping[socket_name]
             for socket_name in connected_inputs[component_name]
         }
+        # TODO special case: leaf nodes should write to state
```
elijahbenizzy, line 360: This isn't clear -- mind clarifying what this means?

elijahbenizzy (commented):
Looks good, some nites

elijahbenizzy (approved):

Issue #417, opened 2024-11-06
fix: Improve error message for `SQLitePersister`


zilto:
I always forget about calling `.initialize()` on the persister to create tables and that's the error I get: 

```python
  File .../assistant.py", line 86, in build_assistant
    .build()
     ^^^^^^^
  File ".../burr/telemetry.py", line 276, in wrapped_fn
    return call_fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../application.py", line 2333, in build
    self._load_from_persister()
  File ".../burr/core/application.py", line 2262, in _load_from_persister
    load_result = self.initializer.load(_partition_key, _app_id, _sequence_id)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../burr/core/persistence.py", line 242, in load
    cursor.execute(
sqlite3.OperationalError: no such table: burr_state
```

Notes:
- it's a bit odd that the stack trace bounces between `user code -> burr.telemetry -> user code -> burr.core.persistence`
- could catch the uninitialized persister at `ApplicationBuilder().with_state_persistence(...)` instead of `ApplicationBuilder().build()`. (seems like a unlikely pattern to pass the persister then initialize it later)
- error message should say `RuntimeError: Uninitialized persister. Make sure to call `.initialize()` before passing it to the `ApplicationBuilder`

arpitgupta-it:
@zilto can I take it up?

zilto:
@arpitgupta-it Yes, you can work on it! Do you have any question? The goal is to catch more explicitly the missing table when calling `self.initializer.load(...)` and raise an appropriate error message

arpitgupta-it:
I just raised a PR. Please review if it looks good to approve.

skrawcz:
@arpitgupta-it thanks I just added comments. We need to extend it to cover all persisters and maintain backwards compatibility with custom persisters people may have written.

PR #418, opened 2024-11-08
fix: Improve error message for SQLitePersister #417

arpitgupta-it:

## Checklist

----

> [!IMPORTANT]
> Add initialization check for persisters in `with_state_persister` to prevent runtime errors with uninitialized persisters.
> 
>   - **Behavior**:
>     - Add check in `with_state_persister` in `application.py` to raise `RuntimeError` if `persister` is not initialized.
>   - **Persistence**:
>     - Add `_initialized` attribute and `is_initialized()` method to `SQLLitePersister` in `persistence.py` to track initialization state.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=DAGWorks-Inc%2Fburr&utm_source=github&utm_medium=referral)<sup> for 5e31502d42e0bf3f71d0251168d7d67fb76f2cfc. It will automatically update as commits are pushed.</sup>


burr/core/application.py
```diff
@@ -2207,6 +2207,10 @@ def with_state_persister(
         """
         if on_every != "step":
             raise ValueError(f"on_every {on_every} not supported")
+        if hasattr(persister, 'initialize') and not persister.is_initialized():
```
skrawcz, line 2210: this is a good idea, but we need to add this to the persister interface and propagate it to the other persisters. To be backwards compatible with we would need to check for the existence of `is_initialized()` and only use it if available -- and therefore only error if it exists, else default to the old behavior of not checking.

skrawcz (changes requested):

burr/core/persistence.py
```diff
@@ -88,7 +88,11 @@ class BaseStatePersister(BaseStateLoader, BaseStateSaver, metaclass=ABCMeta):
     Extend this class if you want an easy way to implement custom state storage.
     """
 
-    pass
+    def is_initialized(self) -> bool:
+        """Check if the persister has been initialized. Default behavior is False.
+        Persisters that require initialization can override this method.
+        """
+        return False
```
skrawcz, line 95: this should raise a `NotImplementedError`.

Then we need the subclasses of this that are in Burr to implement this -- and much like the SQLLite set some internal value.

burr/core/application.py
```diff
@@ -2207,10 +2207,14 @@ def with_state_persister(
...
-            "Uninitialized persister. Make sure to call .initialize() before passing it to the ApplicationBuilder."
-            )
+        
+        # Check if 'is_initialized' exists and whether it returns False, indicating the persister is uninitialized
+        if hasattr(persister, 'is_initialized'):
+            if not persister.is_initialized():
+                raise RuntimeError(
+                    "RuntimeError: Uninitialized persister. Make sure to call .initialize() before passing it to the ApplicationBuilder"
+                )
+        # If the persister is valid and initialized, add it to lifecycle adapters
```
skrawcz, line 2217: Sorry I think I was incorrect. 

1. Everyone should be inheriting from the base. So even custom persisters will have the `is_initialized` method.
2. Instead we want to catch the `not implementederror`.  

So something like this:

```suggestion
        try:
            if not persister.is_initialized():
                raise RuntimeError(
                    "RuntimeError: Uninitialized persister. Make sure to call .initialize() before passing it to the ApplicationBuilder"
                )
        except NotImplementedError:
             pass
        # If the persister is valid and initialized (or hasn't implemented `is_initialized()`), add it to lifecycle adapters
```

skrawcz: we should then have a unit test for this case where a persister has not implemented the method.

skrawcz (commented):
on the right track, some more comments!

arpitgupta-it:
Suggested changed added!

skrawcz (commented):
I think things look good. Asking @elijahbenizzy for review.

burr/core/application.py
```diff
@@ -2207,9 +2207,19 @@ def with_state_persister(
         """
         if on_every != "step":
             raise ValueError(f"on_every {on_every} not supported")
+
         if not isinstance(persister, persistence.BaseStateSaver):
             self.lifecycle_adapters.append(persister)
         else:
+            # Check if 'is_initialized' exists and is False; raise RuntimeError, else continue if not implemented
+            try:
```
elijahbenizzy, line 2215: I think this would be cleaner if it has the attribute `is_initialized` (if `hasattr(persister, "is_initialized")`), or something like that...

burr/core/persistence.py
```diff
@@ -54,6 +54,10 @@ def initialize(self):
         """Initializes the app for saving, set up any databases, etc.. you want to here."""
         pass
 
+    def is_initialized(self) -> bool:
```
elijahbenizzy, line 57: Oh, I see, we're catching the `NotImplemented`  as before...

Might be simpler just to have this default to `True`, then allow you to overwrite it.

burr/core/application.py
```diff
@@ -2207,9 +2207,19 @@ def with_state_persister(
...
+                if not persister.is_initialized():
+                    raise RuntimeError(
```
elijahbenizzy, line 2217: Might be nice to have a subclass `UnintiailizedPersisterException`, or just use `ValueError` -- while it's a bit debated, `RuntimeError` is generally a state that's specifically not supposed to happen: https://discuss.python.org/t/meaning-of-runtimeerror/56206

burr/core/persistence.py
```diff
@@ -192,6 +195,17 @@ def initialize(self):
         """Creates the table if it doesn't exist"""
         # Usage
         self.create_table_if_not_exists(self.table_name)
+        self._initialized = True
+
+    def is_initialized(self) -> bool:
```
elijahbenizzy, line 200: MIght be nice to have a comment about why this is backwards compatible/how it works (E.G. why you don't need to call initialize).

burr/core/persistence.py
```diff
@@ -192,6 +195,17 @@ def initialize(self):
...
+        """Check if persister is properly initialized."""
+        if self._initialized:
+            return True
+        cursor = self.connection.cursor()
+        cursor.execute(
+            "SELECT name FROM sqlite_master WHERE type='table' AND name=?", (self.table_name,)
+        )
+        return cursor.fetchone() is not None
```
elijahbenizzy, line 208: You could also set `self._initialized` to `True` here -- then maybe reuse the same one? E.G. cache it? It'll never go from initialized to uninitialized...

elijahbenizzy (commented):
Looking good! Thank you! Left a few comments -- should be easy to fix up.

burr/core/application.py
arpitgupta-it: Yes, we're implementing `NotImplemented` to tackle this.

burr/core/persistence.py
skrawcz: No because that leads to edge cases with backwards compatibility.

burr/core/persistence.py
skrawcz: no you do need to call initialize() at some point, but not in the lifecycle of the object, since it's a DB with state we can check whether things are correct or not. But sure.

burr/core/persistence.py
```diff
@@ -192,6 +195,17 @@ def initialize(self):
         """Creates the table if it doesn't exist"""
         # Usage
         self.create_table_if_not_exists(self.table_name)
+        self._initialized = True
+
+    def is_initialized(self) -> bool:
+        """Check if persister is properly initialized."""
```
skrawcz, line 201:
```suggestion
        """Check if persister is properly initialized. If the object state is set, we skip checking directly, else we query the DB."""
```

burr/core/persistence.py
skrawcz: oh sure, yeah we should do that.

burr/integrations/persisters/postgresql.py
```diff
@@ -115,14 +116,26 @@ def create_table(self, table_name: str):
...
+    def is_initialized(self) -> bool:
+        """Returns .initialized() has been called"""
+        if self._initialized:
+            return True
+        cursor = self.connection.cursor()
+        cursor.execute(
+            "SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = %s)",
+            (self.table_name,),
+        )
+        return cursor.fetchone()[0]
```
skrawcz, line 130: ditto set initialized value.

burr/core/application.py
skrawcz: I'm in this camp:

> For me, raising RuntimeError is for serious failures of implementation.
It isn’t for situations like raising ValueError, when I’ve been given a
bad value. It’s for situation which should never occur, particularly
things which imply a bad decision or action earlier.

arpitgupta-it:
Hi @elijahbenizzy and @skrawcz,

I’ve reviewed the feedback and the discussions above, but I’m a bit unsure about the specific changes that still need to be made. Could you please clarify what needs to be updated or adjusted in the PR based on the comments so far?

I want to make sure everything is aligned before moving forward with the changes.

Thanks!

burr/core/application.py
elijahbenizzy: Yeah, I read that as more of an internal thing, E.G. not user-facing. The user has made a mistake by not initializing here -- more of a bad state error.

skrawcz:
> Hi @elijahbenizzy and @skrawcz,
> 
> I’ve reviewed the feedback and the discussions above, but I’m a bit unsure about the specific changes that still need to be made. Could you please clarify what needs to be updated or adjusted in the PR based on the comments so far?
> 
> I want to make sure everything is aligned before moving forward with the changes.
> 
> Thanks!

They were very minor things -- you can see my commit for them; thought that would be easier than explaining. I think we should be good to go after these changes.

arpitgupta-it:
Thanks for handling the updates @skrawcz ! Looks like we’re all set... 🤞

elijahbenizzy (approved):
Looks good!

PR #428, opened 2024-11-20
examples: Burr + Hamilton

zilto:
Narrative example which presents the 2-layer approach and shows the typical progression of a RAG or LLM-agent application


skrawcz (approved):
:shipit:

PR #430, opened 2024-11-20
notebook integration: `%burr_ui` can show the web app in an `iframe`

zilto:
Adds the arguments `--no-iframe` (default is enabling iframe) and `--height`.

Notes:
- it doesn't seem to make sense to resize to less than `100%` so there's not `width` arg
- calling `%burr_ui` multiple times will check if the subprocess exited before retrying. If there's a bug in launching the UI, the user may repeatedly hit "launch"
- there's no way to kill the subprocess except restarting the notebook kernel. However,  we guard against launching more than 1 subprocess.

### DIFF
--- a/burr/integrations/notebook.py
+++ b/burr/integrations/notebook.py
@@ -120,7 +120,8 @@ class NotebookMagics(Magics):
 
     @magic_arguments()
     @argument(
-        "--height", "-h",
+        "--height",
+        "-h",
         default=400,
         type=int,
         help="Height of the Burr UI iframe specified as a number of pixels",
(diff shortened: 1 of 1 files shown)


skrawcz (approved):

burr/integrations/notebook.py
```diff
@@ -115,15 +118,35 @@ def __init__(self, notebook_env: NotebookEnvironment, **kwargs):
...
+        else:
+            # if .poll() is not None, then subprocess exited. Try launching the server again
+            # TODO investigate `.returncode` for better failure/retry handling
+            if self.process.poll() is not None:
+                self.process, self.url = launch_ui(self.notebook_env)
+
+        if args.no_iframe is True:
+            print(f"Burr UI: {self.url}")
+        else:
+            display(IFrame(self.url, width="100%", height=args.height))
```
skrawcz, line 149: So the default is to show it, but you can disable it?

skrawcz (approved):

skrawcz:
@zilto just need to re-run pre-commit and fix what it wants so it passes.

PR #435, opened 2024-11-23
Wires through persistence to spawned children apps

elijahbenizzy:
TODO:

- [x] Add documentation
- [x] Add customization for this
  - [x] The ability to specify persisters cascading through -- loader, saver, or both
  - [x] The ability to specify custom persistence capabilities for sub-applications
- [x] Copying of persisters so we don't have the same one
- [ ] Ensuring serialization works (building on (3))
- [x] Add tests

## Checklist

----

> [!IMPORTANT]
> Enhances parallel task execution by adding persistence capabilities to child applications, supporting both sync and async operations, and includes comprehensive tests.
> 
>   - **Behavior**:
>     - Adds persistence capabilities to child applications spawned in parallel.
>     - Supports both synchronous and asynchronous operations.
>     - Introduces `BaseCopyable` interface for cloning objects.
>   - **Functions**:
>     - Adds `asyncify_generator()` and `arealize()` in `async_utils.py` to handle sync and async generators.
>     - Implements `_cascade_adapter()` in `parallelism.py` to manage adapter behavior.
>   - **Classes**:
>     - Adds `BaseCopyable` in `types.py` and `persistence.py` for object cloning.
>     - Updates `ApplicationContext` in `application.py` to include `state_initializer` and `state_persister`.
>     - Introduces `SQLitePersister` and `InMemoryPersister` in `persistence.py` for state management.
>   - **Tests**:
>     - Adds tests for `async_utils.py` to verify generator handling.
>     - Includes tests for persistence in `test_persistence.py` and parallelism in `test_parallelism.py`.
>     - End-to-end tests in `test_end_to_end.py` for parallel task execution with persistence.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=DAGWorks-Inc%2Fburr&utm_source=github&utm_medium=referral)<sup> for 152c355de7628b3c77088464b092c578837db2ef. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/burr/common/async_utils.py
+++ b/burr/common/async_utils.py
@@ -1,0 +1,40 @@
+import inspect
+from typing import AsyncGenerator, AsyncIterable, Generator, List, TypeVar, Union
+
+T = TypeVar("T")
+
+GenType = TypeVar("GenType")
+ReturnType = TypeVar("ReturnType")
+
+SyncOrAsyncIterable = Union[AsyncIterable[T], List[T]]
+SyncOrAsyncGenerator = Union[Generator[GenType, None, None], AsyncGenerator[GenType, None]]
+SyncOrAsyncGeneratorOrItemOrList = Union[SyncOrAsyncGenerator[GenType], List[GenType], GenType]
+
+
+async def asyncify_generator(
+    generator: SyncOrAsyncGenerator[GenType],
+) -> AsyncGenerator[GenType, None]:
+    """Convert a sync generator to an async generator.
+
+    :param generator: sync generator
+    :return: async generator
+    """
+    if inspect.isasyncgen(generator):
+        async for item in generator:
+            yield item
+    else:
+        for item in generator:
+            yield item
+
+
+async def arealize(maybe_async_generator: SyncOrAsyncGenerator[GenType]) -> List[GenType]:
+    """Realize an async generator or async iterable to a list.
+
+    :param maybe_async_generator: async generator or async iterable
+    :return: list
+    """
+    if inspect.isasyncgen(maybe_async_generator):
+        out = [item async for item in maybe_async_generator]
+    else:
+        out = [item for item in maybe_async_generator]
+    return out
--- a/burr/common/async_utils.py
+++ b/burr/common/async_utils.py
@@ -4,7 +4,6 @@ from typing import AsyncGenerator, AsyncIterable, Generator, List, TypeVar, Unio
 T = TypeVar("T")
 
 GenType = TypeVar("GenType")
-ReturnType = TypeVar("ReturnType")
 
 SyncOrAsyncIterable = Union[AsyncIterable[T], List[T]]
 SyncOrAsyncGenerator = Union[Generator[GenType, None, None], AsyncGenerator[GenType, None]]
@@ -31,7 +30,7 @@ async def arealize(maybe_async_generator: SyncOrAsyncGenerator[GenType]) -> List
     """Realize an async generator or async iterable to a list.
 
     :param maybe_async_generator: async generator or async iterable
-    :return: list
+    :return: list of items -- fully realized
     """
     if inspect.isasyncgen(maybe_async_generator):
         out = [item async for item in maybe_async_generator]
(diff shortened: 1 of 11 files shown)


skrawcz:
it seems like the map state etc classes should also be able to override or specify whether they inherit the persister or not. Is that taken into account here?

E.g. default behavior is inherit. Otherwise a user can customize the persister if they need to -- e.g. constructor variable when creating the class?

elijahbenizzy:
> it seems like the map state etc classes should also be able to override or specify whether they inherit the persister or not. Is that taken into account here?
> 
> 
> 
> E.g. default behavior is inherit. Otherwise a user can customize the persister if they need to -- e.g. constructor variable when creating the class?

This is a draft -- see TODOs in commits it's up next

elijahbenizzy:
OK, have not written the documentation but the API is as follows:

On either `MapStates`, `MapActions`, `MapActionsAndStates`, you can specify the following methods:
- `state_persister` -- returns either `"cascade"`, `None`, or a `BaseStatePersister` object
- `state_initializer` -- ditto, but with `BaseStateLoader`
- `tracker` -- ditto, but. with a `TrackingClient` object

for each of these (adapter means tracker/persister)
- `None` means that the base applications do not get the adapter
- `"cascade"` means that the spawning parent's adapter is copied + sent to base applications (tracker is responsible for serialization)
- an instantiated object means that that object is directly sent to the base applications

Note that these are across all tasks -- they will attempt to be copied if specified, although that should be togglable.

Furthermore, you can also customize behavior with the `SubGraphTask` class, which allows you to specify the same fields above (`tracker`, `state_persister`, `state_initializer`) as fields in the constructor.

The default behavior, across the board, is `cascade` -- this is backwards compatible with tracking (which, by default, cascades), and storage should be the same.

skrawcz:
> OK, have not written the documentation but the API is as follows:
> 
> On either `MapStates`, `MapActions`, `MapActionsAndStates`, you can specify the following methods:
> 
> * `state_persister` -- returns either `"cascade"`, `None`, or a `BaseStatePersister` object
> * `state_initializer` -- ditto, but with `BaseStateLoader`
> * `tracker` -- ditto, but. with a `TrackingClient` object
> 
My concern is hard coding this in classes. A user should be able to make this a parameterization of the application. E.g. via the constructor of the action; I'm not sure an input level parameterization makes sense (yet). So as long as we should how to do that and make the action agnostic I think this approach works.

elijahbenizzy:
> > OK, have not written the documentation but the API is as follows:
> > On either `MapStates`, `MapActions`, `MapActionsAndStates`, you can specify the following methods:
> > 
> > * `state_persister` -- returns either `"cascade"`, `None`, or a `BaseStatePersister` object
> > * `state_initializer` -- ditto, but with `BaseStateLoader`
> > * `tracker` -- ditto, but. with a `TrackingClient` object
> 
> My concern is hard coding this in classes. A user should be able to make this a parameterization of the application. E.g. via the constructor of the action; I'm not sure an input level parameterization makes sense (yet). So as long as we show how to do that and make the action agnostic I think this approach works.

Yes those are methods of the action class that is subclassed so what you want to do is easy. 

I anticipate the persister/tracker as a property of the action, not a property of the state/application, so the higher-level API does not consume the state/inputs to create it. They have `**kwargs` exposed for future backwards compatibility.

elijahbenizzy:
Another thoguht I had based on a converstaion with @skrawcz -- in many cases the state_loader and state_initializer is the same. That said, it's not in *all* cases.

My take is that, if the parent's is the same (most cases), and the behavior is set to `"cascade"` the child should be the same as well. Otherwise it'll create two separate instances. Fine to be different, but it's cleaner to add that check. Will implement then look over a use-case or two. 

Use-cases of different versus the same versus one supplied andwhat not supplied:
- both supplied, same: standard persistence/checkpointing
- initializer supplied, persister not -- loading up from production to debug -- cascading is good here
- persister supplied, initializer not -- not sure -- this is weird
- neither supplied -- not persisting

So my take is that the initializer is first and the persister wants to be the same *if* they're the same...

skrawcz (approved):

PR #448, opened 2024-12-03
Demos class-based action

elijahbenizzy:
WIP


elijahbenizzy:
#449 gets at the underlying issue

PR #449, opened 2024-12-03
Fixes __context for classed based actions

skrawcz:
1. Python name mangles __variables on methods.
2. We need to account for that for things to work and inject it properly.

We don't modify the inputs dict since we don't want to change anything, so we shallow copy it, and change the single key value.

Updates docs and adds example to show people.

## Changes
 - tweak to running class based actions
 - docs
 - test

## How I tested this
 - locally via example
 - via unit tests

## Checklist

----

> [!IMPORTANT]
> Fixes handling of `__context` in class-based actions by remapping mangled variable names and updates documentation and tests accordingly.
> 
>   - **Behavior**:
>     - Adds `_remap_context_variable()` in `application.py` to handle name mangling of `__context` in class-based actions.
>     - Updates `_run_function()` in `application.py` to use `_remap_context_variable()` when `__context` is in inputs.
>   - **Documentation**:
>     - Updates `actions.rst` to include examples of using `__context` in class-based actions.
>     - Adds `application_classbased.py` as an example of class-based actions using `__context`.
>   - **Tests**:
>     - Adds tests for `_remap_context_variable()` in `test_application.py` to verify correct remapping of `__context`.
>     - Adds `test_application.py` to test class-based action behavior with `__context`.
>   - **Misc**:
>     - Updates `README.md` in `hello-world-counter` to reference `application_classbased.py` for class-based action examples.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=DAGWorks-Inc%2Fburr&utm_source=github&utm_medium=referral)<sup> for 5a36d46b03cf0dccb7340231eb9e43205c77d791. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/burr/core/application.py
+++ b/burr/core/application.py
@@ -103,6 +103,31 @@ def _adjust_single_step_output(
     _raise_fn_return_validation_error(output, action_name)
 
 
+def _remap_context_variable(run_method: Callable, inputs: Dict[str, Any]) -> dict:
+    """This is a utility function to remap the __context variable to the mangled variable in the function signature.
+
+    Python mangles the variable name in the function signature, so we need to remap it to the correct variable name.
+
+    :param run_method: the run method to inspect.
+    :param inputs: the inputs to inspect
+    :return: potentially new dict with the remapped variable, else the original dict.
+    """
+    # Get the signature of the method being run. This should be Function.run() or similar.
+    signature = inspect.signature(run_method)
+    # Find the name-mangled __context variable
+    mangled_context_name = None
+    for param in signature.parameters.values():
+        if param.name.endswith("__context"):
+            mangled_context_name = param.name
+            break
+
+    # If a mangled __context variable is found, remap the value in inputs
+    if mangled_context_name and "__context" in inputs:
+        inputs = inputs.copy()
+        inputs[mangled_context_name] = inputs.pop("__context")
+    return inputs
+
+
 def _run_function(function: Function, state: State, inputs: Dict[str, Any], name: str) -> dict:
     """Runs a function, returning the result of running the function.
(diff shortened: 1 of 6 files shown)


burr/core/application.py
```diff
@@ -103,6 +103,31 @@ def _adjust_single_step_output(
     _raise_fn_return_validation_error(output, action_name)
 
 
+def _remap_context_variable(run_method: Callable, inputs: Dict[str, Any]) -> dict:
```
elijahbenizzy, line 106: We should do this with any `__` variable... `__tracer` will hit the same problem. Could hardcode this, but we should consider being more aware of the context.

burr/core/application.py
```diff
@@ -121,6 +146,9 @@ def _run_function(function: Function, state: State, inputs: Dict[str, Any], name
         )
     state_to_use = state.subset(*function.reads)
     function.validate_inputs(inputs)
+    if "__context" in inputs:
+        # potentially need to remap the __context variable
+        inputs = _remap_context_variable(function.run, inputs)
```
elijahbenizzy, line 151: I don't think remapping the context variable should be aware of the signature of the run function. Who knows how it would be created...

Rather, let's just hardcode the name-mangled version and unmangle them? Or maybe do it upstream?

elijahbenizzy (commented):
Hmm, I think the change should be made here? https://github.com/DAGWorks-Inc/burr/blob/1cde57e0601c665cc4af440e44e7e81409ddd104/burr/core/action.py#L553

We should be able to fix the issue by how we inspect it...

skrawcz:
will handle `__tracer` later in another commit/PR.

elijahbenizzy (approved):

Issue #484, opened 2024-12-20
Async persister interface


elijahbenizzy:
**Is your feature request related to a problem? Please describe.**
The current [persister interface](https://github.com/DAGWorks-Inc/burr/blob/50d33522b609774cd4c916a68131dc44b7844c5c/burr/core/persistence.py#L98) is synchronous. This can cause performance issues in asynchronous code.

**Describe the solution you'd like**
We should have a separate stack for asynchronous persistence -- ideally with modes for either (1) fire-and-forget or (2) blocking/transactional (depends on how consistent the user wants it -- this part might be general persister configuration, rather than async-specific).

To dig in, we will need:
- [ ] Persistence interfaces for this (`BaseStateLoader`, `BaseStateSaver`, `PersisterHook`) -- each converted to async
- [ ] A few default implementations (SQLLite, redis? TBD -- go with the easy ones on this)
- [ ] Integrations with `application`
    - [ ] asynchronous `build()` method in application builder (`.abuild()`) - or maybe another builder? Guess is we want an asynchronous build method
    - [ ] validation in `.run()` -- E.G. if you have async hooks they will not be called -- at least a warning message if not an error
- [ ] Comprehensive testing for ^^^
- [ ] Think through integration with parallel actions  --- will want an `acreate_app` [here](https://github.com/DAGWorks-Inc/burr/blob/50d33522b609774cd4c916a68131dc44b7844c5c/burr/core/parallelism.py#L113).

**Describe alternatives you've considered**

Could have synchronous fire/forget mode, although we very likely want something transactional.

elijahbenizzy:
This is done, thanks @jernejfrank

PR #488, opened 2024-12-27
Async persistance

jernejfrank:
Addressing #484 to have an async persistence interface.

## Changes
- Adds base classes for saving/loading/running persistor in application via async
- Adds async initializer/persister to ApplicationBuilder and lets us build async Application via `.abuild()`
- Adds async validation to Application to warn about async hook being ignored in sync runs and raises error if you try to do sync run with an async Application.
- Adds create_async_app in parallelism to create application via AsyncApplicationBuilder (in case of legacy use of sync initializer/persister reverts to old sync Application)
- TBD: database specific implementations (only AsyncDevNullPersister for testing)

To think about: (1) fire-and-forget or (2) blocking/transactional options -- 
Async persistors are naturally (1) and sync persistors are (2). If we want to have both options for both cases, we need to:
- effectively block the async save to turn (1)->(2) and 
- create an event loop / another tread and make sync save a coroutine executing there to go from (2) -> (1).  
My initial idea was to have that option as an attribute of the persistor class and handle it within PersisterHook/PersisterHookAsync, but still checking if there is a more natural place for this.

## How I tested this
- Unit and E2E

## Notes
This part is very much for discussion if we want to have another `AsyncApplicationBuilder` or put this into the existing `ApplicationBuilder`. My arguments for having two:
1. It is cleaner to separate async and sync (and maybe less error-prone to use the wrong one?).
2. There are 3 methods we need to define async: `with_state_persister`, `_load_from_persister`, and `build`. 
3. Related to the above, if `with_state_persister` is async it gets a bit hairy how to do method chaining -- what I did is to overwrite the original method to just store the state_persister (similar to what `initialize_from` does) and then pushed all the logic into an async helper function `__with_async_state_persister` that gets awaited in `build` to manually chain coroutines.
4. Having another builder class made it clearer also in the Application class to raise an error when `run` is used instead of `arun`.

Having said all of that, I also can push those methods into the existing builder with an `abuild()` to follow the pattern in the Application class where both sync and async are side-by-side.

## Checklist

- [] New functions are documented (with a description, list of inputs, and expected output)

----

> [!IMPORTANT]
> Add asynchronous persistence support with `AsyncApplicationBuilder` and related async classes and tests.
> 
>   - **Behavior**:
>     - Introduces `AsyncApplicationBuilder` to support asynchronous state persistence.
>     - Adds `create_async_app` in `parallelism.py` for async application creation.
>     - Implements `AsyncDevNullPersister` for testing async persistence.
>   - **Persistence**:
>     - Adds `AsyncBaseStateLoader` and `AsyncBaseStateSaver` in `persistence.py` for async state operations.
>     - Introduces `PersisterHookAsync` for async lifecycle hooks.
>   - **Testing**:
>     - Adds tests for async persistence in `test_application.py`, including `test_async_save_and_load_from_persister_end_to_end`.
>   - **Misc**:
>     - Updates imports in `__init__.py` and `application.py` to include async classes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=DAGWorks-Inc%2Fburr&utm_source=github&utm_medium=referral)<sup> for 9e233c887482dd04ee96662b3cd8bc739f8f1a72. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/burr/core/application.py
+++ b/burr/core/application.py
@@ -44,7 +44,12 @@ from burr.core.action import (
     StreamingResultContainer,
 )
 from burr.core.graph import Graph, GraphBuilder
-from burr.core.persistence import BaseStateLoader, BaseStateSaver
+from burr.core.persistence import (
+    AsyncBaseStateLoader,
+    AsyncBaseStateSaver,
+    BaseStateLoader,
+    BaseStateSaver,
+)
 from burr.core.state import State
 from burr.core.typing import ActionSchema, DictBasedTypingSystem, TypingSystem
 from burr.core.validation import BASE_ERROR_MESSAGE
@@ -87,7 +92,9 @@ def _raise_fn_return_validation_error(output: Any, action_name: str):
 
 
 def _adjust_single_step_output(
-    output: Union[State, Tuple[dict, State]], action_name: str, action_schema: ActionSchema
+    output: Union[State, Tuple[dict, State]],
+    action_name: str,
+    action_schema: ActionSchema,
 ):
     """Adjusts the output of a single step action to be a tuple of (result, state) or just state"""
 
@@ -839,6 +846,7 @@ class Application(Generic[ApplicationStateType]):
         """
         # we need to increment the sequence before we start computing
         # that way if we're replaying from state, we don't get stuck
+        self.validate_correct_async_use()
         self._increment_sequence_id()
         out = self._step(inputs=inputs, _run_hooks=True)
         return out
@@ -1037,7 +1045,10 @@ class Application(Generic[ApplicationStateType]):
                     )
                 else:
                     result = await _arun_function(
-                        next_action, self._state, inputs=action_inputs, name=next_action.name
+                        next_action,
+                        self._state,
+                        inputs=action_inputs,
+                        name=next_action.name,
                     )
                     new_state = _run_reducer(next_action, self._state, result, next_action.name)
                 new_state = self._update_internal_state_value(new_state, next_action)
@@ -1174,6 +1185,7 @@ class Application(Generic[ApplicationStateType]):
         :return: Each iteration returns the result of running `step`. This generator also returns a tuple of
             [action, result, current state]
         """
+        self.validate_correct_async_use()
         halt_before, halt_after, inputs = self._clean_iterate_params(
             halt_before, halt_after, inputs
         )
@@ -1237,6 +1249,7 @@ class Application(Generic[ApplicationStateType]):
             Note that this is only used for the first iteration -- subsequent iterations will not use this.
         :return: The final state, and the results of running the actions in the order that they were specified.
         """
+        self.validate_correct_async_use()
         gen = self.iterate(halt_before=halt_before, halt_after=halt_after, inputs=inputs)
         while True:
             try:
@@ -1382,6 +1395,7 @@ class Application(Generic[ApplicationStateType]):
                 result, state = output.get()
                 print(format(result['response'], color))
         """
+        self.validate_correct_async_use()
         call_execute_method_wrapper = _call_execute_method_pre_post(ExecuteMethod.stream_result)
         call_execute_method_wrapper.call_pre(self)
         halt_before, halt_after, inputs = self._clean_iterate_params(
@@ -1957,6 +1971,29 @@ class Application(Generic[ApplicationStateType]):
         dot = self.visualize(include_conditions=True, include_state=False)
         return dot._repr_mimebundle_(include=include, exclude=exclude, **kwargs)
 
+    def validate_correct_async_use(self):
+        """Validates that the application is meant to run async.
+        This validation is performed in synchronous application methods."""
+
+        # This is a gentle warning for existing users to use the async application
+        if self._adapter_set.async_hooks:
+            logger.warning(
+                "There are asynchronous hooks present in the application that will be ignored. "
+                "Please use async methods to run the application and have them executed. "
+                f"The application has following asynchronous hooks: {self._adapter_set.async_hooks} "
+            )
+
+        # We check that if:
+        # - we have build the application using .abuild()
+        # - we have async hooks present
+        # this application is meant to be run in async mode.
+        if self._builder and self._builder.is_async:
+            raise ValueError(
+                "The application was build with .abuild() "
+                "which needs to be executed in an asynchronous run. "
+                "Please use the async run methods to run the application."
+            )
+
 
 def _validate_app_id(app_id: Optional[str]):
     if app_id is None:
@@ -1999,6 +2036,7 @@ class ApplicationBuilder(Generic[StateType]):
         self.typing_system = None
         self.parallel_executor_factory = None
         self.state_persister = None
+        self._is_async: bool = False
 
     def with_identifiers(
         self, app_id: str = None, partition_key: str = None, sequence_id: int = None
@@ -2150,7 +2188,9 @@ class ApplicationBuilder(Generic[StateType]):
         return self
 
     def with_actions(
-        self, *action_list: Union[Action, Callable], **action_dict: Union[Action, Callable]
+        self,
+        *action_list: Union[Action, Callable],
+        **action_dict: Union[Action, Callable],
     ) -> "ApplicationBuilder[StateType]":
         """Adds an action to the application. The actions are granted names (using the with_name)
         method post-adding, using the kw argument. If it already has a name (or you wish to use the function name, raw, and
@@ -2168,7 +2208,8 @@ class ApplicationBuilder(Generic[StateType]):
     def with_transitions(
         self,
         *transitions: Union[
-            Tuple[Union[str, list[str]], str], Tuple[Union[str, list[str]], str, Condition]
+            Tuple[Union[str, list[str]], str],
+            Tuple[Union[str, list[str]], str, Condition],
         ],
     ) -> "ApplicationBuilder[StateType]":
         """Adds transitions to the application. Transitions are specified as tuples of either:
@@ -2249,7 +2290,7 @@ class ApplicationBuilder(Generic[StateType]):
 
     def initialize_from(
         self,
-        initializer: BaseStateLoader,
+        initializer: Union[BaseStateLoader, AsyncBaseStateLoader],
         resume_at_next_action: bool,
         default_state: dict,
         default_entrypoint: str,
@@ -2295,7 +2336,9 @@ class ApplicationBuilder(Generic[StateType]):
         return self
 
     def with_state_persister(
-        self, persister: Union[BaseStateSaver, LifecycleAdapter], on_every: str = "step"
+        self,
+        persister: Union[BaseStateSaver, AsyncBaseStateSaver, LifecycleAdapter],
+        on_every: str = "step",
     ) -> "ApplicationBuilder[StateType]":
         """Adds a state persister to the application. This is a way to persist state out to a database, file, etc...
         at the specified interval. This is one of two options:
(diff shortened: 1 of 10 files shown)


elijahbenizzy (commented):
Lookking good -- some broad comments. I'm falling on the side of making it a single builder to force sharing of code , IMO it makes it a bit simpler + duplicates less logic. That said, i see both ways of doing it.

About this:

> To think about: (1) fire-and-forget or (2) blocking/transactional options --
> Async persistors are naturally (1) and sync persistors are (2). If we want to have both options for both cases, we need to:
> 
> effectively block the async save to turn (1)->(2) and
> create an event loop / another tread and make sync save a coroutine executing there to go from (2) -> (1).
> My initial idea was to have that option as an attribute of the persistor class and handle it within PersisterHook/PersisterHookAsync, but still checking if there is a more natural place for this.

To clarify -- async persisters are *not* naturally fire-and-forget -- they'll await completion of the task, which blocks the event loop. `asyncio.ensure_future` (I think that's the right one) is a way to make it fire/forget. You're right about the thread -- we'll need to tie into some sort of execution. This is also why we can probably push this up in the stack -- E.G. we don't use the result of these for the next execution, so we can put it in a queue.

We would have to worry about how to ensure order, especially in persisters (there's probably an async queue pattern that works -- should be pretty easy, but definitely more complicated). This also gets to some interesting db/distributed systems problems -- I'm not convinced there aren't other ugly pitfalls here.

So I think we should keep it transactional as it is now...

burr/core/application.py
```diff
@@ -1174,6 +1184,22 @@ def iterate(
...
+        # This is a gentle warning for existing users
+        if self._adapter_set.async_hooks:
+            logger.warning(
+                "There are asynchronous hooks present in the application that will be ignored. "
+                "Please use .aiterate or .arun methods to have them executed. "
+                f"The application has following asynchronous hooks: {self._adapter_set.async_hooks} "
+            )
+
+        # Seems fair to raise this if everything is async but the app execution
+        if self._adapter_set.async_hooks and isinstance(self._builder, AsyncApplicationBuilder):
```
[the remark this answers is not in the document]
jernejfrank: IMO we should error out when the application is built with `.abuild()`, has async hooks, and we run it with sync methods, e.g. `.run()` since this is very easy to miss that the async hooks just don't get called.

Vice versa, if we have an async application with `.arun()` and sync hooks, I wouldn't error it out since we cannot guarantee that we every adapter has async support / user might be ok for some things to be blocking. Maybe a warning here is more appropriate?

burr/core/application.py
```diff
@@ -1466,8 +1492,9 @@ def callback(
                     sequence_id=self.sequence_id,
                     exception=None,
                 )
-                out = action, StreamingResultContainer.pass_through(
-                    results=result, final_state=state
+                out = (
+                    action,
+                    StreamingResultContainer.pass_through(results=result, final_state=state),
```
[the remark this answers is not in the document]
jernejfrank: No, might have been my local settings..

burr/core/persistence.py
```diff
@@ -94,13 +126,67 @@ def save(
...
+        :param app_id: Appliaction UID to write with
+        :param sequence_id: Sequence ID of the last executed step
+        :param position: The action name that was implemented
+        :param state: The current state of the application
+        :param status: The status of this state, either "completed" or "failed". If "failed" the state is what it was
+            before the action was applied.
+        """
+        pass
+
+    def is_async(self) -> bool:
```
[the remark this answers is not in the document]
jernejfrank: It is used now in the `ApplicationBuilder` to validate :)

burr/core/application.py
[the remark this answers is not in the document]
elijahbenizzy: Yes, I agree -- this is backwards compatible. The second case is odd, actually, as you've pointed out. To make it more concrete, not every synchronous hook is incompatible in an async mode. Take, for example, the hook that prints "doing X step" before every step -- this is just going to be a sync hook, not an async hook, but it should work in an async setting. No need to build an async version of something like that, when it never has an await.

I think we shold maintain the design where `.arun()` allows sync + async hooks, `.run()` only allows sync hooks (and should break otherwise, althugh that's an edge case...). Also, think this simplifies?

burr/core/application.py
```diff
@@ -2484,3 +2589,51 @@ def build(self) -> Application[StateType]:
...
+        """
+        # setting this here so that we can differentiate which builder was used in the Application
+        self._is_async = True
+
+        _validate_app_id(self.app_id)
+        if self.state is None:
+            self.state = State()
+
+        if self.state_persister:
+            await self._set_async_state_persister()  # this is used to save the state during application run
```
elijahbenizzy, line 2629: I'm thinking that `.abuild()` should allow a sync persister but log a warning. But maybe not? Seems fair to force it here...

elijahbenizzy (commented):
Looks really good -- as noted in a DM I think we have to align on behavior. E.G. when using `.build()` versus `.abuild()`, sync versus async persister, `amethod` versus `method` -- which configurations break/log warning/work? Cartesian product/table is a clean way to think of it.

jernejfrank:
> Looks really good -- as noted in a DM I think we have to align on behavior. E.G. when using `.build()` versus `.abuild()`, sync versus async persister, `amethod` versus `method` -- which configurations break/log warning/work? Cartesian product/table is a clean way to think of it.

screenshot_2024-12-29_at_17 32 31

This is a current working proposition of which cases should allow what. I think we can replace `persister` with `hooks` and put it in the builder docs as well to have clarity on how to use `.abuild()`.

elijahbenizzy:
> > Looks really good -- as noted in a DM I think we have to align on behavior. E.G. when using `.build()` versus `.abuild()`, sync versus async persister, `amethod` versus `method` -- which configurations break/log warning/work? Cartesian product/table is a clean way to think of it.
> 
> screenshot_2024-12-29_at_17 32 31
> 
> This is a current working proposition of which cases should allow what. I think we can replace `persister` with `hooks` and put it in the builder docs as well to have clarity on how to use `.abuild()`.



> > Looks really good -- as noted in a DM I think we have to align on behavior. E.G. when using `.build()` versus `.abuild()`, sync versus async persister, `amethod` versus `method` -- which configurations break/log warning/work? Cartesian product/table is a clean way to think of it.
> 
> screenshot_2024-12-29_at_17 32 31
> 
> This is a current working proposition of which cases should allow what. I think we can replace `persister` with `hooks` and put it in the builder docs as well to have clarity on how to use `.abuild()`.

OK, yeah, I think this is the right way to do it! Nice work enumerating it, you have me convinced. The only case (and I'm not sure how this works in the one above) is when you have a synchronous hook in an async `.abuild()` method. Not a persister (so not fully covered by the above), but I think it *should* work. That said, we should have appropriate documentation about it.

burr/core/application.py
```diff
@@ -1957,6 +1971,24 @@ def _repr_mimebundle_(self, include=None, exclude=None, **kwargs):
...
+        # This is a gentle warning for existing users
+        if self._adapter_set.async_hooks:
+            logger.warning(
+                "There are asynchronous hooks present in the application that will be ignored. "
+                "Please use async methods to run the application and have them executed. "
+                f"The application has following asynchronous hooks: {self._adapter_set.async_hooks} "
+            )
+
+        # Seems fair to raise this if everything is async but the app execution is sync
+        if self._adapter_set.async_hooks and self._builder.is_async:
```
elijahbenizzy, line 1985: Confused -- this is erroring out if the builder is async and the adapter has async hooks? `.is_async`?

burr/core/application.py
```diff
@@ -2484,3 +2589,51 @@ def build(self) -> Application[StateType]:
             state_persister=self.state_persister,
             state_initializer=self.state_initializer,
         )
+
+    @telemetry.capture_function_usage
```
elijahbenizzy, line 2593: Nice this function is extremely clean

burr/core/application.py
elijahbenizzy: Scratch that, yes, that case should not work

burr/core/application.py
```diff
@@ -2484,3 +2589,51 @@ def build(self) -> Application[StateType]:
...
+            # sets state, sequence_id, and maybe start
+            self._load_from_sync_persister()  # this is used to load application from a previously saved state
+
+        return self._build_common()
+
+    @telemetry.capture_function_usage
+    async def abuild(self) -> Application[StateType]:
+        """Builds the application.
+
+        This function is a bit messy as we iron out the exact logic and rigor we want around things.
```
elijahbenizzy, line 2617: No longer messy, fix this. And make it + `build()` clear about the parameters (good place to have this table?).

burr/core/persistence.py
```diff
@@ -148,6 +261,33 @@ def save(
...
+        self,
+        partition_key: Optional[str],
+        app_id: str,
+        sequence_id: int,
+        position: str,
+        state: State,
+        status: Literal["completed", "failed"],
+        **kwargs,
+    ):
+        # print("I saved something.")
```
elijahbenizzy, line 287: remove

burr/core/parallelism.py
```diff
@@ -133,8 +135,59 @@ def run(
...
+            builder = builder.initialize_from(
+                self.state_initializer,
+                default_state=self.state.get_all(),  # TODO _- ensure that any hidden variables aren't used...
+                default_entrypoint=self.graph.entrypoint,
+                resume_at_next_action=True,
+            )
+        else:
+            builder = builder.with_entrypoint(self.graph.entrypoint).with_state(self.state)
+
+        return await builder.abuild()
```
elijahbenizzy, line 173: OK, complexity, this has the off-chance of making something backwards incompatible. That said, I think that's OK TBH. In this case it's an odd one:
1. Someone is using the async parallelism piece
2. Someone is *not* using the async persister (which they won't be, it isn't there)
3. It breaks on reloading

I think very few people are doing this, but it *might* be worth having a parameter in `.abuild()` to bypass validation, we could do that here...

elijahbenizzy (commented):
Looking good! Some small points:
1. I think we should document this clearly. Ideally is a `sync` versus `async` page in concepts, but this could just be part of the docstring
2. A bit of a strange case in parallelism, where we call the new function by default. Might be worth adding in a parameter to bypass validation. Or just breaking -- it's an edge case. 

Otherwise, let's ship soon! Nice work!

burr/core/application.py
jernejfrank: Yes! So I am calling this in the sync methods and it  covers cases 3 and 7 in the above table. Thinking back, maybe it's better to be strict from the get go and only check `_builder.is_async` and error out.

Previously, if there were no async hooks present (i.e. no async persisters, etc.) it doesn't really matter how the app was built or is run -- since only sync hooks in the app. But might as well just say if you build it async, use it async.

burr/core/parallelism.py
jernejfrank: We can add the parameter, but not sure if it is that helpful. If they are using sync persisters they will have to go through the sync `.build()` method anyway since `.abuild()` expects async persisters.

This is what I capture now  parallelism `.arun()`: we check if we have sync persisters and then use `.build()` to have that backward compatibility, otherwise everything async and ok.

jernejfrank:
> OK, yeah, I think this is the right way to do it! Nice work enumerating it, you have me convinced. The only case (and I'm not sure how this works in the one above) is when you have a synchronous hook in an async `.abuild()` method. Not a persister (so not fully covered by the above), but I think it _should_ work. That said, we should have appropriate documentation about it.

This works as before. You run both sync and async hooks --> sync ones are blocking. Also, if you use sync persister with `.build()` it will still let you run async (this is backwards compatibility). But if you use the new `.abuild()` method, it complains in there about using sync persisters and later when you have the application it complains that you are trying to run it sync.

burr/core/application.py
```diff
@@ -1957,6 +1971,29 @@ def _repr_mimebundle_(self, include=None, exclude=None, **kwargs):
...
+                f"The application has following asynchronous hooks: {self._adapter_set.async_hooks} "
+            )
+
+        # We check that if:
+        # - we have build the application using .abuild()
+        # - we have async hooks present
+        # this application is meant to be run in async mode.
+        if self._builder and self._builder.is_async:
+            raise ValueError(
+                "The application was build with async hooks "
```
elijahbenizzy, line 1992: Make this say "it was built with `abuild`" -- think that's clearer/more actionable

docs/concepts/sync-vs-async.rst
```diff
@@ -0,0 +1,117 @@
+===========================
+Sync vs Async Applications
+===========================
+
+TL;DR
```
elijahbenizzy, line 5: Link to the `arun`, `aiterate`, etc... methods. Good to tie it back into the actions.

elijahbenizzy (approved):
Nice work! Really big improvements. Left some small comments but it's there.

PR #504, opened 2025-01-24
Fixes state writes with private variables

elijahbenizzy:
This drops private variables when written by an action. Note we do *not* add tests for this as it is undefined behavior -- we may error out later on.

----

> [!IMPORTANT]
> `_state_update` in `application.py` now drops private variables when written by an action, with a warning added in `state.rst` about undefined behavior.
> 
>   - **Behavior**:
>     - `_state_update` in `application.py` now drops private variables (keys starting with `__`) when written by an action.
>     - Modifying private variables is considered undefined behavior and may result in errors in the future.
>   - **Documentation**:
>     - Added warning in `state.rst` about undefined behavior when modifying private variables starting with `__`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=DAGWorks-Inc%2Fburr&utm_source=github&utm_medium=referral)<sup> for 00f17f4f36a60d01ed8fee7bc73f3d966bea4b4a. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/burr/core/application.py
+++ b/burr/core/application.py
@@ -180,17 +180,11 @@ async def _arun_function(
 
 
 def _state_update(state_to_modify: State, modified_state: State) -> State:
-    """This is a hack to apply state updates and ensure that we are respecting deletions. Specifically, the process is:
-
-    1. We subset the state to what we want to read
-    2. We perform a set of state-specific writes to it
-    3. We measure which ones were deleted
-    4. We then merge the whole state back in
-    5. We then delete the keys that were deleted
+    """Performs a state update for an action -- allowing for deletes.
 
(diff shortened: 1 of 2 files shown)


skrawcz (commented):
LGTM. but are you sure there's no test to be written? At least to break if we modify the current behavior?

skrawcz (approved):

elijahbenizzy:
> LGTM. but are you sure there's no test to be written? At least to break if we modify the current behavior?

Yes, it's specifically undefined (as in we want to address it further but not now), so I'm comfortable not adding tests for now as to not encode a contract we don't need.

Issue #468, opened 2024-12-11
Tags for actions


elijahbenizzy:
**Is your feature request related to a problem? Please describe.**
We should be able to tag actions. In particular, this could be used to do halt_after and halt_before.

**Describe the solution you'd like**

```python
@action(reads=[...], writes=[...], tags=["requires_human_input"])
def act(state: State, human_input: ...) -> State:
    ...

app.run(halt_before=["requires_human_input"])
```

Would have to do this at the class-level as well, should be straightforward. Also there's the possibility of name-clashes on halt. I'm OK with that due to the simplicity here.

Could also do this at the action level -- in the builder `with_actions(some_action=action.with_tags(["requires_human_input"])`. Would have to think about how to do this with function-based actions (TBD).

**Describe alternatives you've considered**
See above

**Additional context**
Something I've been mulling over for a while.

skrawcz:
suggestion to know how to read it better:
***@***.***:requires_human_input"])

elijahbenizzy:
OK, thoughts:

- these are node name aliases
- Stopping before/after any of X,Y,Z is the use-case — this is a common problem I’ve observed where people enumerate out node names, hard-coded. Tags form aliases for multiple nodes. `@action(tags=[“foo”, “bar”]) ->  halt_after=["foo", "bar"], mixed freely with ["node_1", "node_2"])
- Will work for class/function-based (although quite a few APIs to cover here) — likely an instance method
(pending time) Will likely have system-level tags (E.G. __has__input and __has_input__<foo>) for named inputs — the non-dynamically-generated ones will be constants, otherwise it will be a function
    - This specifically solves the problem of pausing before all inputs and all inputs of form X (E.G. multiple need the same)
    -IMO this is key for human in the loop/control flow.
- Tags will have very specific naming structures — python var name, no starting with underscore, no ending with undrescore

elijahbenizzy:
This is complete! See #505

PR #505, opened 2025-01-26
Adds action tagging

elijahbenizzy:
This enables us to use tags as aliases for actions. We may add future tags (E.G. __requires_inputs), but for now this is all done on behalf of the user. The idea is to have a very simple polymorphic interface -- we can (for example) say all nodes tagged with "text_return" fullfill the same role (displaying markdown for the user).

See #468 for more details.

## Changes

- [x] Implements tagging interface
- [x] Adds unit/integration tests
- [x] Docs
    - [x] ensure docstrings
    - [x] concepts/whatnot
- [ ] E2E test (at least one)

## How I tested this
Tests + manual.

## Checklist

----

> [!IMPORTANT]
> Introduces action tagging to group actions by aliases, updating core classes, tests, and documentation.
> 
>   - **Behavior**:
>     - Introduces action tagging, allowing actions to be tagged with aliases for grouping and identification.
>     - Updates `Application` class to process control flow parameters with tags in `_process_control_flow_params()`.
>     - Supports tags in `halt_before` and `halt_after` parameters in `Application` methods.
>   - **Graph**:
>     - Adds `_create_action_tag_map()` to map tags to actions in `Graph`.
>     - Implements `get_actions_by_tag()` to retrieve actions by tag.
>   - **Pydantic Integration**:
>     - Adds `tags` parameter to `pydantic_action()` and `pydantic_streaming_action()`.
>   - **Tests**:
>     - Adds tests for action tagging in `test_action.py`, `test_application.py`, and `test_graph.py`.
>   - **Documentation**:
>     - Updates `actions.rst` to include information on action tagging.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=DAGWorks-Inc%2Fburr&utm_source=github&utm_medium=referral)<sup> for 4123a9e7a78d0baa84847b98972d03a8632c0cda. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/burr/core/action.py
+++ b/burr/core/action.py
@@ -239,6 +239,15 @@ class Action(Function, Reducer, abc.ABC):
         write_repr = ", ".join(self.writes) if self.writes else "{}"
         return f"{self.name}: {read_repr} -> {write_repr}"
 
+    @property
+    def tags(self) -> list[str]:
+        """Returns the tags associated with this action.
+        Tags are effectively action aliases -- names that apply towards multiple actions.
+
+        :return: List of string tags
+        """
+        return []
+
 
 class Condition(Function):
     KEY = "PROCEED"
@@ -586,6 +595,7 @@ class FunctionBasedAction(SingleStepAction):
         input_spec: Optional[tuple[list[str], list[str]]] = None,
         originating_fn: Optional[Callable] = None,
         schema: ActionSchema = DEFAULT_SCHEMA,
+        tags: Optional[List[str]] = None,
     ):
         """Instantiates a function-based action with the given function, reads, and writes.
         The function must take in a state and return a tuple of (result, new_state).
@@ -611,6 +621,7 @@ class FunctionBasedAction(SingleStepAction):
             )
         )
         self._schema = schema
+        self._tags = tags if tags is not None else []
 
     @property
     def fn(self) -> Callable:
@@ -632,6 +643,10 @@ class FunctionBasedAction(SingleStepAction):
     def schema(self) -> ActionSchema:
         return self._schema
 
+    @property
+    def tags(self) -> list[str]:
+        return self._tags
+
     def with_params(self, **kwargs: Any) -> "FunctionBasedAction":
         """Binds parameters to the function.
         Note that there is no reason to call this by the user. This *could*
@@ -649,6 +664,7 @@ class FunctionBasedAction(SingleStepAction):
             input_spec=self._inputs,
             originating_fn=self._originating_fn,
             schema=self._schema,
+            tags=self._tags,
         )
 
     def run_and_update(self, state: State, **run_kwargs) -> tuple[dict, State]:
@@ -1063,6 +1079,7 @@ class FunctionBasedStreamingAction(SingleStepStreamingAction):
         input_spec: Optional[tuple[list[str], list[str]]] = None,
         originating_fn: Optional[Callable] = None,
         schema: ActionSchema = DEFAULT_SCHEMA,
+        tags: Optional[List[str]] = None,
     ):
         """Instantiates a function-based streaming action with the given function, reads, and writes.
         The function must take in a state (and inputs) and return a generator of (result, new_state).
@@ -1086,6 +1103,7 @@ class FunctionBasedStreamingAction(SingleStepStreamingAction):
         )
         self._originating_fn = originating_fn if originating_fn is not None else fn
         self._schema = schema
+        self._tags = tags if tags is not None else []
 
     async def _a_stream_run_and_update(
         self, state: State, **run_kwargs
@@ -1115,6 +1133,10 @@ class FunctionBasedStreamingAction(SingleStepStreamingAction):
     def streaming(self) -> bool:
         return True
 
+    @property
+    def tags(self) -> list[str]:
+        return self._tags
+
     def with_params(self, **kwargs: Any) -> "FunctionBasedStreamingAction":
         """Binds parameters to the function. This is not user-facing -- this is
         meant to be used internally by the API.
@@ -1130,6 +1152,7 @@ class FunctionBasedStreamingAction(SingleStepStreamingAction):
             input_spec=self._inputs,
             originating_fn=self._originating_fn,
             schema=self._schema,
+            tags=self._tags,
         )
 
     @property
@@ -1204,6 +1227,7 @@ class action:
         writes: List[str],
         state_input_type: Optional[Type["BaseModel"]] = None,
         state_output_type: Optional[Type["BaseModel"]] = None,
+        tags: Optional[List[str]] = None,
     ) -> Callable:
         """Action that specifies inputs/outputs using pydantic models.
         This should make it easier to develop with guardrails.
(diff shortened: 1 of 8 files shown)


burr/core/action.py
```diff
@@ -239,6 +239,15 @@ def __repr__(self):
         write_repr = ", ".join(self.writes) if self.writes else "{}"
         return f"{self.name}: {read_repr} -> {write_repr}"
 
+    @property
+    def tags(self) -> list[str]:
+        """Returns the tags associated with this action.
+        Tags are effectively action aliases -- names that apply towards multiple actions.
```
skrawcz, line 245: can this be applied to edges too?

skrawcz: on thinking more applying to edges is fraught - so should mention where this alias is to be used?

skrawcz (changes requested):
Can we get the UX of how someone uses this scoped first?

I'm of the strong opinion we should demarcate what is a tag and what isn't in the control flow. E.g.

```python
 ...  = app.run(halt_before=["@tag:display_inputs"], inputs=...)
```

With a straight up alias I see these issues otherwise:

1. What if an action has the same name?
2. When reading the control flow you don't know what it refers to easily without knowing the actions themselves.

So requirements thoughts:

1. When someone reads the control flow it should be clear that tags are special. Since tags could map to multiple nodes. Hence why I like a prefix.
2. You should be able to have an action name and tag be the same thing, you'd distinguish them with one being `@tag:` or something. 
3. From an edge perspective, you could see tags functioning on the "source" side of the edge to expand that and route to a single node, e.g. error node.
4. You should be able to get the mapping of tags -> actions from the graph/application object.
5. I don't see an error case, but with the above an action can have multiple tags and things should all just work.
6. We might want a tag object, that way people don't need to update strings, and instead can use the pointer to that object... (though we need this more broadly to be able to replace "strings" when defining a graph/application).

elijahbenizzy:
> Can we get the UX of how someone uses this scoped first?
> 
> I'm of the strong opinion we should demarcate what is a tag and what isn't in the control flow. E.g.
> 
> ```python
>  ...  = app.run(halt_before=["@tag:display_inputs"], inputs=...)
> ```
> 
> With a straight up alias I see these issues otherwise:
> 
> 1. What if an action has the same name?
> 2. When reading the control flow you don't know what it refers to easily without knowing the actions themselves.
> 
> So requirements thoughts:
> 
> 1. When someone reads the control flow it should be clear that tags are special. Since tags could map to multiple nodes. Hence why I like a prefix.
> 2. You should be able to have an action name and tag be the same thing, you'd distinguish them with one being `@tag:` or something.
> 3. From an edge perspective, you could see tags functioning on the "source" side of the edge to expand that and route to a single node, e.g. error node.
> 4. You should be able to get the mapping of tags -> actions from the graph/application object.
> 5. I don't see an error case, but with the above an action can have multiple tags and things should all just work.
> 6. We might want a tag object, that way people don't need to update strings, and instead can use the pointer to that object... (though we need this more broadly to be able to replace "strings" when defining a graph/application).

Yeah, thought through it a bit. Trade-offs:

1. On one hand you've got a bespoke UX -- `@tag:` -- is this a common pattern? Are there others (`@requires_input:`? `@streaming_response:`)? feels like it'll likely be a one-off, althought requires_input is OK. It's also string magic
2. On the other hand it's readable -- clear what's happening.

My thought was to make them indistinguishable from each other -- if they have the right tag then they effectively function the same. I'm not sure the value in distinguishing them -- if we have it so names are true aliases, then it's very clear that two nodes are interchangeable. 

That said `@tag` does improve clarity, but I wanted to keep the API simple. Regarding an action with the same name -- that's a user error, but there's an interesting point -- tags are set at the action-level whereas names are set at the application level (often with defaults at the action-level)... One could also add tags at the application level as well `.with_tags(...)`, which we have not implemented yet.

So yeah, I'm not against `"@tag:<tag_name>"` even though it's quite a bit uglier -- I think the clarity might be worthwhile worthwhile.

skrawcz:
> 1. On one hand you've got a bespoke UX -- `@tag:` -- is this a common pattern? Are there others (`@requires_input:`? `@streaming_response:`)? feels like it'll likely be a one-off, althought requires_input is OK. It's also string magic

Those are properties of the action though? They could be property. E.g. `@property:requires_input` `@property:streaming_response`.

> My thought was to make them indistinguishable from each other -- if they have the right tag then they effectively function the same. I'm not sure the value in distinguishing them -- if we have it so names are true aliases, then it's very clear that two nodes are interchangeable.

I don't think that's the problem being solved here. To me it's:

1. Not having to manually list all actions.
2. Being able to divorce function/class name from that of provided action name to the application. I.e. one less thing to update if you change the names of things. Tags would be invariant here. (though likely need a way to remove tags for edge cases here).

We err on the side of making things more readable, and to try to help a user mentally map what is going on / is different from just an "action name". E.g. `@tag:...` could correspond to 1 or more actions, which isn't true the other way around.

elijahbenizzy:
> > 1. On one hand you've got a bespoke UX -- `@tag:` -- is this a common pattern? Are there others (`@requires_input:`? `@streaming_response:`)? feels like it'll likely be a one-off, althought requires_input is OK. It's also string magic
> 
> Those are properties of the action though? They could be property. E.g. `@property:requires_input` `@property:streaming_response`.
> 
> > My thought was to make them indistinguishable from each other -- if they have the right tag then they effectively function the same. I'm not sure the value in distinguishing them -- if we have it so names are true aliases, then it's very clear that two nodes are interchangeable.
> 
> I don't think that's the problem being solved here. To me it's:
> 
> 1. Not having to manually list all actions.
> 2. Being able to divorce function/class name from that of provided action name to the application. I.e. one less thing to update if you change the names of things. Tags would be invariant here. (though likely need a way to remove tags for edge cases here).
> 
> We err on the side of making things more readable, and to try to help a user mentally map what is going on / is different from just an "action name". E.g. `@tag:...` could correspond to 1 or more actions, which isn't true the other way around. So I think it would be something hard to walk back if we allowed it to look like a regular action.

So (1) is the goal here -- (2) I'd rephrase as "being able to divorce the actual instance of the action from the stated goal (E.G. polymorphism), allowing easier iteration. We don't use the class name, and the function name is just a convenience.

So I'm ok with `@tag:...` for readability, but also want to add `.with_tags(...)` I think.

elijahbenizzy:
@skrawcz and I talked offline -- I think `@tag:...` is probably better, added that. Will test then merge tomorrow.

burr/core/application.py
```diff
@@ -1575,7 +1601,7 @@ async def astream_result(
         which will be empty. Thus ``halt_after`` takes precedence -- if it is met, the streaming result container will contain the result of the
         halt_after condition.
 
-        The :py:class:`AsyncStreamingResultContainer <burr.core.action.StreamingResultContainer>` is meant as a convenience -- specifically this allows for
+        The :py:class:`StreamingResultContainer <burr.core.action.StreamingResultContainer>` is meant as a convenience -- specifically this allows for
```
skrawcz, line 1604: ?

skrawcz (approved):

PR #522, opened 2025-03-10
S3 Tracker Logging - Excessive Logging Removal

mdrideout:
The S3 tracker has excessive logging.

## Changes

- Removed excessive logging from S3 tracker. Added clarification to existing flushing logs.

## How I tested this

Tested the branch directly with

```toml
burr = { git = "https://github.com/mdrideout/burr.git", branch = "s3-logging", extras = [
    "graphviz",
    "opentelemetry",
    "tracking-client-s3",
] }
```

## Checklist


burr/tracking/s3client.py
```diff
@@ -215,7 +215,6 @@ def thread(self):
 
         while self.running:
             try:
-                logger.info(f"Checking for new data to flush -- batch is of size: {len(batch)}")
```
skrawcz, line 218: we could switch this to debug mode, and check if in debug mode or not.

skrawcz:
@mdrideout question, is INFO level useful at all? Maybe they should all move to DEBUG? Or some split between that?

mdrideout:
If there is a debug mode that could be useful. My objective was to make it easier for us to debug our backend, and this specific log seemed like it was leftover from development. 

Our issue is that the excessive logging makes our server logs and dev instance terminal very cluttered since it runs every 1-3 seconds.

skrawcz:
So I would propose moving them to `logger.debug`.
You can then optionally turn on debug logging if you want to see those logs or not.

burr/tracking/s3client.py
mdrideout: Branch updated to bring back the original log, and convert these to `.debug` logs.

mdrideout:
Branch updated to bring back the original log, and convert these to .debug logs.

skrawcz (approved):

skrawcz:
thanks @mdrideout !

PR #535, opened 2025-05-06
Pool based async persister

gamarin2:
This PR adds a way to instantiate the async postgres persister with a connection pool. The goal is to prevent a bug in parallelism: when the user implements async versions of map-reduce classes (i.e. set is_async to True), burr calls `asyncio.gather` instead of threads, meaning we can't reuse the connection of the async persister. 

One solution is to update `b_asyncpg` to use a connection pool instead of a single connection and add a copy method to it, meaning that when burr cascades to get persisters for each sub-app, we get a new connection from the pool each time. 

## Changes

- Added a way to instantiate `AsyncPostgresPersister` with a pool instead of a connection
- Added a copy method that favors using the pool in copied instances if self.pool is not None
- Kept everything backwards compatible
- Also added missing async related docs

## How I tested this

Manual testing

## Notes

So far I only implemented a proposed solution in `b_asyncpg`, but the issue is still present in other async persister implementations. Once we agree on a definitive solution, I can update them as well. 

Open question:

- Do we really want to keep things backwards compatible and let user instantiate with a simple connection? In most scenarios a pool is safer, especially in async context

## Checklist

----

> [!IMPORTANT]
> Add connection pool support to `AsyncPostgreSQLPersister` for better parallelism handling in async contexts.
> 
>   - **Behavior**:
>     - `AsyncPostgreSQLPersister` can now be instantiated with a connection pool using `from_values()` with `use_pool=True`.
>     - Adds `copy()` method to `AsyncPostgreSQLPersister` to create new instances using the pool.
>     - Updates `_get_connection()` and `_release_connection()` to handle pooled connections.
>   - **Documentation**:
>     - Updates `actions.rst`, `parallelism.rst`, and `sync-vs-async.rst` to include information about using connection pools with async persisters.
>   - **Misc**:
>     - Maintains backward compatibility with direct connections.
>     - Fixes minor formatting issues in `b_asyncpg.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=DAGWorks-Inc%2Fburr&utm_source=github&utm_medium=referral)<sup> for c4d6293bb3c9dadbd9ea387fab6e5b1687167aaf. You can [customize](https://app.ellipsis.dev/DAGWorks-Inc/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/burr/integrations/persisters/b_asyncpg.py
+++ b/burr/integrations/persisters/b_asyncpg.py
@@ -25,6 +25,11 @@ class AsyncPostgreSQLPersister(persistence.AsyncBaseStatePersister, BaseCopyable
         We suggest to use the persister either as a context manager through the ``async with`` clause or
         using the method ``.cleanup()``.
 
+    .. warning::
+        If you intend to use parallelism features or need to share this persister across multiple tasks,
+        you should initialize it with a connection pool (set ``use_pool=True`` in ``from_values``).
+        Direct connections cannot be shared across different tasks and may cause errors in concurrent scenarios.
+
     .. note::
         The implementation relies on the popular asyncpg library: https://github.com/MagicStack/asyncpg
 
--- a/burr/integrations/persisters/b_asyncpg.py
+++ b/burr/integrations/persisters/b_asyncpg.py
@@ -1,16 +1,22 @@
-from burr.integrations import base
+import json
+import logging
+from typing import Literal, Optional, ClassVar
+from typing import Any
 from burr.common.types import BaseCopyable
+from burr.core import persistence, state
+from burr.integrations import base
+
 
 try:
     import asyncpg
 except ImportError as e:
     base.require_plugin(e, "asyncpg")
 
-import json
-import logging
-from typing import Literal, Optional, ClassVar
+try:
+    from typing import Self
+except ImportError:
+    Self = Any
 
-from burr.core import persistence, state
 
 logger = logging.getLogger(__name__)
 
@@ -41,10 +47,10 @@ class AsyncPostgreSQLPersister(persistence.AsyncBaseStatePersister, BaseCopyable
 
     .. code:: bash
 
-        docker run --name local-psql \\  # container name
-                   -v local_psql_data:/SOME/FILE_PATH/ \\  # mounting a volume for data persistence
-                   -p 54320:5432 \\  # port mapping
-                   -e POSTGRES_PASSWORD=my_password \\  # superuser password
+        docker run --name local-psql \\\\  # container name
+                   -v local_psql_data:/SOME/FILE_PATH/ \\\\  # mounting a volume for data persistence
+                   -p 54320:5432 \\\\  # port mapping
+                   -e POSTGRES_PASSWORD=my_password \\\\  # superuser password
                    -d postgres  # database name
 
     Then you should be able to create the class like this:
--- a/burr/integrations/persisters/b_asyncpg.py
+++ b/burr/integrations/persisters/b_asyncpg.py
@@ -1,4 +1,5 @@
 from burr.integrations import base
+from burr.common.types import BaseCopyable
 
 try:
     import asyncpg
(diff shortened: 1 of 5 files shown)


skrawcz (commented):
This looks good to me. Will defer to @elijahbenizzy since I wasn't too close to the async persister stuff.

Also seems like our pre-commit hook doesn't like a couple of things...

```
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted burr/core/parallelism.py

All done! ✨ 🍰 ✨
1 file reformatted, 203 files left unchanged.

trim trailing whitespace.................................................Passed
fix end of files.........................................................Failed
- hook id: end-of-file-fixer
- exit code: 1
- files were modified by this hook

Fixing docs/concepts/parallelism.rst

fix requirements.txt.....................................................Passed
check python ast.........................................................Passed
isort....................................................................Failed
- hook id: isort
- files were modified by this hook

Fixing /home/runner/work/burr/burr/burr/integrations/persisters/b_asyncpg.py

flake8...................................................................Failed
- hook id: flake8
- exit code: 1

<unknown>:24: SyntaxWarning: invalid escape sequence '\ '
<unknown>:84: SyntaxWarning: invalid escape sequence '\-'
burr/integrations/persisters/b_asyncpg.py:154:23: F821 undefined name 'Self'
```

elijahbenizzy (approved):
This looks great! Only thing missing is testing, LMK if you want to add it, but otherwise just add a TODO. I think this one is tricky to tests TBH.

gamarin2:
I pushed fixes for the pre-commit hooks. I'd still like to know your opinion on the following @elijahbenizzy : I have kept things backwards compatible, but should we really let people instantiate async persisters with direct connections? This means people that use this will still run into the issue when using parallelism. In async setting, I would argue it'd be better to "force" connection pools to be used

Regarding tests, there are already some in the test folder that I can draw inspiration from, but if you have something specific in mind lmk

elijahbenizzy:
> I pushed fixes for the pre-commit hooks. I'd still like to know your opinion on the following @elijahbenizzy : I have kept things backwards compatible, but should we really let people instantiate async persisters with direct connections? This means people that use this will still run into the issue when using parallelism. In async setting, I would argue it'd be better to "force" connection pools to be used
> 
> Regarding tests, there are already some in the test folder that I can draw inspiration from, but if you have something specific in mind lmk

@gamarin2 Ok, I'm not sure this is "backwards compatible" as much as "preserves prior behavior". E.G. will anyone's code break if this is released and the default is set to `True` rather than `False` for `use_pool`?

I don't think so, although you have a better sense. 

If not: let's set the default to the optimal behavior and save people a headache.
If so: let's think more carefully about this -- we're still in `0.X.X` so we have the right, but we want to be smart about this. 

@jernejfrank -- any thoughts?

jernejfrank (commented):
Hey @gamarin2 , great work and thanks for opening the PR! I left some minor comments.

@elijahbenizzy RE default setting:

I don't think opening a connection pool instead of a connection will go terribly wrong for existing stuff. The one place I could see it go wrong is some multithreading happening outside of Burr and then opening the connection pool would result in a deadlock. 

+1 from me to make the default using the pool (as it is the more sensible option for Burr).

burr/integrations/persisters/b_asyncpg.py
```diff
@@ -48,11 +60,35 @@ class AsyncPostgreSQLPersister(persistence.AsyncBaseStatePersister):
...
+        if cls._pool is None:
+            cls._pool = await asyncpg.create_pool(
+                user=user,
+                password=password,
+                database=database,
+                host=host,
+                port=port,
+                **pool_kwargs,
+            )
+        return cls._pool
```
asgillmor, line 90: is the idea with the class variable here that it is a singleton object that we reference in `from_values` as a factory?

kind of seems like we should be using _pool in the copy method?

jernejfrank (approved):

elijahbenizzy:
Mergine despite build failures @jernejfrank has agreed to fix!

PR #553, opened 2025-06-20
fix the order of 'context' and 'inputs' arguments in class MapActions' function signature of "actions" 

tiejunhu:
The order of context and inputs was wrong.

## Changes

The actions function signature and the example in docstring

## Checklist


elijahbenizzy (approved):
Thank you!

PR #600, opened 2025-11-28
Hack: manually get around symlink resolution of flit

skrawcz:
Flit doesn't resolve symlinks, so in the build process for the wheel we copy over the files we want. We then build the wheel. Afterwards we reset the symlink.

This now helps ensure that the convenience package has everything it needs to start the server without issue.

## Changes
 - changes build artifact helper

## How I tested this
 - validates that in the wheel all the correct files are present
 - ran the installed wheel to validate

## Checklist

### DIFF
--- a/scripts/build_artifacts.py
+++ b/scripts/build_artifacts.py
@@ -136,16 +136,215 @@ def _clean_ui_build():
         print("✓ UI build directory cleaned.\n")
 
 
+def _replace_symlinks_with_copies():
+    """
+    Replace symlinked example files/directories with actual copies before building.
+    Returns a dict mapping paths to their symlink targets (if they were symlinks),
+    so they can be restored later.
+    """
+    # Files and directories from pyproject.toml lines 266-270 that might be symlinks
+    example_paths = [
+        "examples/__init__.py",
+        "examples/email-assistant",
+        "examples/multi-modal-chatbot",
+        "examples/streaming-fastapi",
+        "examples/deep-researcher",
+    ]
+
+    symlink_info = {}  # Maps path -> (was_symlink: bool, symlink_target: str or None, is_dir: bool)
+
+    for path in example_paths:
+        if not os.path.exists(path):
+            continue
+
+        if os.path.islink(path):
+            # It's a symlink - we need to replace it with a copy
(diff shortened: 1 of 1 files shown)


potiuk:
You might instead opt to use hatchling. In airflow we use "force-include" feature for the very same purpose:

https://github.com/apache/airflow/blob/main/airflow-core/pyproject.toml#L223C1-L227C85

```toml
[tool.hatch.build.targets.sdist.force-include]
"../shared/configuration/src/airflow_shared/configuration" = "src/airflow/_shared/configuration"
"../shared/logging/src/airflow_shared/logging" = "src/airflow/_shared/logging"
"../shared/secrets_masker/src/airflow_shared/secrets_masker" = "src/airflow/_shared/secrets_masker"
"../shared/timezones/src/airflow_shared/timezones" = "src/airflow/_shared/timezones"
```

potiuk:
That is WAY simpler than custom build script.

potiuk:
All it needs is:

```toml
[build-system]
requires = [
    "hatchling==1.27.0",
    "packaging==25.0",
    "pathspec==0.12.1",
    "pluggy==1.6.0",
    "smmap==5.0.2",
    "tomli==2.2.1; python_version < '3.11'",
    "trove-classifiers==2025.9.11.17",
]
build-backend = "hatchling.build"
```

This will provide build reproducibility for the future. When you set-up dependabot it will automatically upgrade the pinned requirements (we contributed a fix to dependabot to add this feature).

elijahbenizzy:
@potiuk yes I like that a lot. I'm going to look at the difference but I'd like to fix it up post release -- have a bit of momentum and want to get this out to get started with a more regular cadence. Will also look at airflow/processes more generally to see what we can learn!

elijahbenizzy:
I have verified that this works

elijahbenizzy (approved):

potiuk:
> @potiuk yes I like that a lot. I'm going to look at the difference but I'd like to fix it up post release -- have a bit of momentum and want to get this out to get started with a more regular cadence. Will also look at airflow/processes more generally to see what we can learn!

Yep. Makes perfect sense.

Issue #607, opened 2025-12-11
Streaming Event type, type hint, should support union type


skrawcz:
# Current behavior
When it's a union type it complains:
```
union = MyModel1 | MyModel2

@streaming_action.pydantic(
    reads=[...],
    writes=[...],
    state_input_type=..., 
    state_output_type=...
    stream_type=union, # type hint says this is invalid
)
```

because it is typed with :

```
stream_type: Union[Type["BaseModel"], Type[dict]],
```

## Library & System Information
Latest burr

# Expected behavior
It should be fine.

# Additional context
The fix I think is to expand that stream type to handle a union of BaseModel types.

AryanBagade:
@skrawcz if its open for external contributors, can i work on it!

andreahlert:
Labels migrated to new taxonomy (#692).

elijahbenizzy:
So this I think doesn't work because it says it could be either or but the type that is "either of them" doesn't satisfy this. Union[A, B] is not a subtype of A or B. 

Rather it should accept Union types directly -- does it actually detect per-instance rather than assume they're all the same?

ThyTran1402:
Hi,

I can work on this issue if it's not resolved yet.

Thank you!

andreahlert:
> Hi,
> 
> I can work on this issue if it's not resolved yet.
> 
> Thank you!

Hello! Feel free to give a try. 

@Ghraven are you still workin on this? 

Can you elaborate together?

Ghraven:
Thanks for checking in. I'm not actively working on this at the moment, so I don't want to block anyone else from taking it.

@ThyTran1402 please feel free to pick it up. My current understanding is that the fix should focus on accepting/recognizing `Union[...]` annotations directly instead of treating a union as needing to satisfy one concrete member type globally. I'd be happy to review or sanity-check an approach if that helps, but I'm fine stepping back so you can move it forward.

andreahlert:
Thanks @Ghraven. @ThyTran1402 assigned to you :)

ThyTran1402:
Hi @andreahlert,

I was able to find the root cause of this issue that `stream_type` parameter was defined with a restrictive type of annotation: 
On `burr/core/action.py`:
```python
stream_type: Union[Type["BaseModel"], Type[dict]]
```
On `burr/integrations/pydantic.py`:
```python
PartialType = Union[Type[pydantic.BaseModel], Type[dict]]
```
It only accepted: 
[...]

PR #617, opened 2025-12-31
fix(pydantic): accept union type objects for stream_type hints

majiayu000:
## Summary
- treat union type objects as valid stream_type values for pydantic streaming actions
- keep runtime behavior unchanged while relaxing type hints for union-type streams

## Testing
- not run (type-hint only change)

### DIFF
--- a/burr/core/action.py
+++ b/burr/core/action.py
@@ -62,6 +62,18 @@ if TYPE_CHECKING:
     except ImportError:
         pass
 
+try:
+    UnionType = types.UnionType
+except AttributeError:  # py<3.10
+    UnionType = None
+
+if typing.TYPE_CHECKING:
+    StreamingPydanticType = Any
+else:
+    StreamingPydanticType = Union[Type["BaseModel"], Type[dict]]
+    if UnionType is not None:
+        StreamingPydanticType = Union[Type["BaseModel"], Type[dict], UnionType]
+
 
 class Function(abc.ABC):
     """Interface to represent the 'computing' part of an action"""
@@ -1309,7 +1321,7 @@ class streaming_action:
         writes: List[str],
         state_input_type: Type["BaseModel"],
         state_output_type: Type["BaseModel"],
-        stream_type: Union[Type["BaseModel"], Type[dict]],
+        stream_type: "StreamingPydanticType",
         tags: Optional[List[str]] = None,
     ) -> Callable:
         """Creates a streaming action that uses pydantic models.
--- a/burr/integrations/pydantic.py
+++ b/burr/integrations/pydantic.py
@@ -22,6 +22,7 @@ import inspect
 import types
 import typing
 from typing import (
+    Any,
     AsyncGenerator,
     Awaitable,
     Callable,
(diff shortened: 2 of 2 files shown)


majiayu000:
Thanks for the review\! I've updated the code according to your suggestion:

- Removed the `TYPE_CHECKING` + `Any` approach
- Added `"types.UnionType"` (with quotes) to the Union type hints

The type definitions are now:
```python
# burr/integrations/pydantic.py
PartialType = Union[Type[pydantic.BaseModel], Type[dict], type, "types.UnionType"]

# burr/core/action.py  
stream_type: Union[Type["BaseModel"], Type[dict], type, "types.UnionType"]
```

This should satisfy pyright while properly accepting union type objects like `str | int` for the `stream_type` parameter.

skrawcz (commented):
Thanks for tackling this! The union type support for `stream_type` is a needed fix for issue #607, and the runtime behavior is correct.

The `TYPE_CHECKING = Any` approach disables type checking entirely — it accepts `stream_type="hello"` without complaint. I'd recommend replacing both `TYPE_CHECKING` blocks with a `sys.version_info >= (3, 10)` guard, which pyright understands natively:

```python
if sys.version_info >= (3, 10):
    StreamingPydanticType = Union[Type[BaseModel], Type[dict], types.UnionType]
else:
    StreamingPydanticType = Union[Type[BaseModel], Type[dict]]
```

This correctly narrows the type on 3.10+ while falling back gracefully on 3.9. I validated this with pyright against all the expected cases (union types, single types, dict, and invalid inputs).

burr/core/action.py
```diff
@@ -62,6 +62,18 @@
...
+    UnionType = types.UnionType
+except AttributeError:  # py<3.10
+    UnionType = None
+
+if typing.TYPE_CHECKING:
+    StreamingPydanticType = Any
+else:
+    StreamingPydanticType = Union[Type["BaseModel"], Type[dict]]
+    if UnionType is not None:
+        StreamingPydanticType = Union[Type["BaseModel"], Type[dict], UnionType]
```
skrawcz, line 75: The `TYPE_CHECKING = Any` pattern effectively disables the type checker for `stream_type`. Replace with:
```python
if sys.version_info >= (3, 10):
    StreamingPydanticType = Union[Type["BaseModel"], Type[dict], types.UnionType]
else:
    StreamingPydanticType = Union[Type["BaseModel"], Type[dict]]
```
`sys` is already imported in this file. Pyright natively narrows on `sys.version_info` checks.

burr/integrations/pydantic.py
```diff
@@ -264,7 +265,17 @@ async def async_action_function(state: State, **kwargs) -> State:
...
+    UnionType = types.UnionType
+except AttributeError:  # py<3.10
+    UnionType = None
+
+if typing.TYPE_CHECKING:
+    PartialType = Any
+else:
+    PartialType = Union[Type[pydantic.BaseModel], Type[dict]]
+    if UnionType is not None:
+        PartialType = Union[Type[pydantic.BaseModel], Type[dict], UnionType]
```
skrawcz, line 278: Same issue — replace the `TYPE_CHECKING = Any` pattern with the `sys.version_info` guard for `PartialType`.

burr/integrations/pydantic.py
```diff
@@ -22,6 +22,7 @@
 import types
 import typing
 from typing import (
```
skrawcz, line 24: Nit: the `Any` import can be removed if switching to the `sys.version_info` approach.

Issue #521, opened 2025-02-25
Update all examples to use latest openai models


skrawcz:
We need to replace old model, e.g. gpt-3.5, etc usage with the latest model equivalents.

We need to search the code base and examples and replace these old models with the latest ones.

skrawcz:
> I would like to work on it if it has not already been assigned or finished.

sure!

goutamk09:
Hi @skrawcz

I see this issue was previously assigned — just wanted to check if it’s still open or if help is welcome.

I have experience working with OpenAI APIs and Python examples, and I’d be happy to:
- Update example code to use the latest OpenAI models
- Align the syntax with the current OpenAI Python client
- Verify that examples still run correctly

Please let me know if I can proceed or assist in any way. Thanks!

skrawcz:
@goutamk09 go ahead.

goutamk09:
Hi @skrawcz, I’ve started by reviewing the OpenAI usage under the `examples/` directory.

I’m beginning with `examples/conversational-rag/graph_db_example/application.py` and updating deprecated model references (e.g. gpt-3.5 / gpt-4) to a current recommended model (`gpt-4o-mini`).  

I’ll proceed incrementally across the remaining example files once this approach looks good. Please let me know if you’d like anything done differently. Thanks!

goutamk09:
Hi @skrawcz   

I’ve opened PR #615 updating the conversational RAG example to use `gpt-4o-mini`.
Happy to continue with the remaining examples once this approach is confirmed. Thanks!

andreahlert:
Labels migrated to new taxonomy (#692).

PR #655, opened 2026-02-10
Update OpenAI model in opentelemetry example

goutamk09:
Updates the OpenAI model used in the opentelemetry example to the latest supported model.

Related to #521


goutamk09:
@skrawcz 
Tested locally after updating the model.
The application initializes and runs; Graphviz rendering error on Windows is expected and unrelated to this change.

PR #656, opened 2026-02-12
core: add AST-based linter for undeclared state reads in function-bas…

### LINKED ISSUE #407: Linter for action reads in state, function-based-actions
I want to be able to quickly fail should I attempt to read the wrong state item.

```python
@action(reads=["foo"], ...)
def my_action(state: State) -> State:
    bar = state["bar"] # not in reads...
    ...
```

We know for sure that this should fail.

A checker to warn this. We won't be able to catch all cases (for whatever reason), but given the following assumptions (that we can check for), we should be able to get it working to be correct.
1. The `state` parameter is not overwritten with another state parameter
2. Only strings in `reads` 
3. State access is done with brackets

Then we can walk through the AST to (a) ensure this and (b) check that all state parameters if that's the case.

Having a better error message, adding that as well.

**Additional context**
Just developing and liked this idea

This is not an easy bit of coding (I'm sure chatGPT can help), but it's self-contained and very testable, so should be doable for most engineers.

Smitaambiger:
Closes #407

This PR adds AST-based validation for undeclared state reads in function-based actions.

### What this does

- Parses the originating function using `inspect.getsource`
- Walks the AST to detect `state["<literal>"]` accesses
- Compares detected keys with declared `reads=[...]`
- Raises `ValueError` if undeclared keys are accessed
- Skips validation when source is unavailable

### Why

Prevents silent bugs caused by accessing state keys that were not declared in `reads`, enforcing stricter action correctness.

### Tests

- Added unit tests covering:
  - Valid declared reads
  - Undeclared read detection


Smitaambiger:
@elijahbenizzy 

Initial implementation for AST-based undeclared state read validation.

Currently implemented as a hard ValueError at action construction time.

Happy to adjust:
- validation timing (build-time vs decorator-time)
- severity (warning vs error)

Looking forward to feedback.

tests/core/test_action_reads_linter.py
skrawcz: you'll also need a test to ensure this doesn't impact pydantic typed state.

Smitaambiger:
@skrawcz ,Thanks for the detailed feedback!

Addressed the requested changes:

- Moved imports to top level
- Consolidated tests into `test_action.py`
- Added test for multiple undeclared keys interleaved with declared keys
- Added regression test to ensure pydantic-based actions are not impacted
- Cleaned up lint/formatting issues

**Core tests are passing locally and CI should reflect the same.**
```
(.venv) PS C:\Users\gouta\OneDrive\Desktop\burr> python -m pytest tests/core -q
................................................................................................................................................................................................. [ 69%]
....................................................................................                                                                                                              [100%]
277 passed in 9.10s
```

```
(.venv) PS C:\Users\gouta\OneDrive\Desktop\burr> pre-commit run --all-files
black....................................................................Passed
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
fix requirements.txt.....................................................Passed
check python ast.........................................................Passed
isort....................................................................Passed
flake8...................................................................Passed
frontend-lint-staged.....................................................Passed
```

Please let me know if any further adjustments are needed.

Smitaambiger:
@skrawcz  Just following up — please let me know if anything else needs adjustment here. Happy to iterate further if needed.

Smitaambiger:
@skrawcz Thanks for the review and guidance — really enjoyed working on this one!

PR #572, opened 2025-08-21
New Graph Builder Tool in Burr UI

jaeyow:
A Graph Builder Tool will allow one to visually create a Burr Graph, by dragging and dropping and interacting on a React Flow-based canvas. This will allow you to view and copy the generated boilerplate Python Burr Graph, which you can then use in your Burr application.

image

## Changes
- added @xyflow/react (React Flow) to enable the graph builder functionality
- added source code editor view (prism-react-renderer) to allow view and copy of generated code to clipboard
- added a new GraphBuilder component with all its supporting functionality 

## How I tested this (all manual tests)
- The graph builder does not depend on any API functionality, so all I had to test was the new Graph Builder functionality in the Burr UI. 
- create, edit and delete Nodes
- create edit and delete Edges
- view and copy generated Python graph to clipboard
- view and copy generated graph JSON to clipboard
- try all the Key commands (keyboard shortcuts)
- expand and collapse left sidepanel
- expand and collapse right sifepanel
- load and view example graphs
- create nodes from te FAB button

## Notes
- a separate pull request is created to refactor the GraphView component to use the same version of ReactFlow as being used in this PR, as this is already getting big. 

## Checklist

- [x] PR has an informative and human-readable title (this will be pulled into the release notes)
- [x] Changes are limited to a single goal (no scope creep)
- [ ] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.
- [x] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future TODOs are captured in comments
- [ ] Project documentation has been updated if adding/changing functionality.

PR #667, opened 2026-03-01
feat: add visual Graph Builder tool to Burr UI

skrawcz:
This takes work from @jaeyow and https://github.com/apache/burr/pull/572.

Adds a drag-and-drop graph editor for designing Burr application graphs visually and exporting as Python code or JSON.

## Changes
* these are UI only changes to add new graph-builder

Key changes:
- New /graph-builder route with full visual editor (ReactFlow v12)
- Migrate existing GraphView from reactflow v11 to @xyflow/react v12
- Remove reactflow and @tisoap/react-flow-smart-edge dependencies
- Per-node async/streaming toggles matching Burr's 4 action variants
- Python code generation with correct decorators and signatures
- 3 pre-built example graphs (MultiModal Chatbot, CRAG, Streaming)
- localStorage auto-save/restore of graph state
- Empty-state onboarding overlay and structured help sidebar
- Fix appcontainer layout for full-height content

## How I tested this
* locally

Screenshot 2026-03-01 at 08 27 49

## Notes

Next would be to unify the color schemes with the observability components.

## Checklist

### DIFF
--- a/telemetry/ui/package-lock.json
+++ b/telemetry/ui/package-lock.json
@@ -12,12 +12,13 @@
         "@heroicons/react": "^2.1.1",
         "@microsoft/fetch-event-source": "^2.0.1",
         "@testing-library/react": "^13.4.0",
-        "@tisoap/react-flow-smart-edge": "^3.0.0",
+        "@tisoap/react-flow-smart-edge": "^4.0.0",
         "@types/node": "^20.0.0",
         "@types/react": "^18.2.56",
         "@types/react-dom": "^18.2.19",
         "@types/react-syntax-highlighter": "^15.5.11",
         "@uiw/react-json-view": "^2.0.0-alpha.12",
+        "@xyflow/react": "^12.0.0",
         "clsx": "^2.1.0",
         "dagre": "^0.8.5",
         "fuse.js": "^7.0.0",
@@ -30,7 +31,6 @@
         "react-router-dom": "^6.22.1",
         "react-select": "^5.8.1",
         "react-syntax-highlighter": "^15.5.0",
-        "reactflow": "^11.10.4",
         "remark-gfm": "^4.0.0",
         "tailwindcss-question-mark": "^0.4.0",
         "typescript": "~6.0.2"
@@ -1431,18 +1431,6 @@
         "node": ">=6.0.0"
       }
     },
-    "node_modules/@jridgewell/source-map": {
-      "version": "0.3.5",
-      "resolved": "https://registry.npmjs.org/@jridgewell/source-map/-/source-map-0.3.5.tgz",
-      "integrity": "sha512-UTYAUj/wviwdsMfzoSJspJxbkH5o1snzwX0//0ENX1u/55kkZZkcTZP6u9bwKGkv+dkk9at4m1Cpt0uY80kcpQ==",
-      "dev": true,
-      "optional": true,
-      "peer": true,
-      "dependencies": {
-        "@jridgewell/gen-mapping": "^0.3.0",
-        "@jridgewell/trace-mapping": "^0.3.9"
-      }
-    },
     "node_modules/@jridgewell/sourcemap-codec": {
       "version": "1.5.5",
       "resolved": "https://registry.npmjs.org/@jridgewell/sourcemap-codec/-/sourcemap-codec-1.5.5.tgz",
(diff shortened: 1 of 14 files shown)


skrawcz (commented):
Nice, self-contained addition — the Graph Builder is cleanly factored (separate exporter / code-generator / loader / presentational components), every new file carries the Apache header, and the generated Python correctly matches Burr's real API including the async-streaming `AsyncGenerator[Tuple[dict, Optional[State]], None]` signature. The `reactflow` v11 → `@xyflow/react` v12 migration looks correct (`edgesUpdatable`→`edgesReconnectable`, named `ReactFlow` import, updated CSS path) and nets a dependency reduction with no impact on the `burr` package, so backwards compatibility is fine.
Before merge I'd like to see:
1. **Unit tests** for `BurrCodeGenerator`, `GraphExporter`, and `ExampleLoader.validateExample` — these are pure functions and the highest-value thing missing. At minimum, assert `validateExample` returns `[]` for each bundled example and snapshot the generated Python for one sync + one streaming graph.
2. Fix the **duplicate-label code-generation bug**: `deduplicateNames` is applied only to the `with_actions` list, not to the generated `def`s or transitions, so two nodes with the same label produce colliding/redefined functions. Dedup consistently or key off `node.id`.
3. A short **docs page** (the PR checklist docs item is unchecked) and a manual smoke-test of existing UI pages given the shared `appcontainer` layout change and the v12 migration.
Also worth a quick confirmation: dropping the smart-edge routing in favor of plain bezier paths in the observability `GraphView` is a minor visual regression (edges may cross nodes) — is that acceptable?

---

- **File:** `telemetry/ui/src/components/routes/graph-builder/utils/BurrCodeGenerator.ts`
  **Line:** ~148 (`generateGraphFunction` — `deduplicateNames` call)
  **[BLOCKING]** `deduplicateNames` is applied here to build the `with_actions(...)` list, but `generateActions` and `generateTransitions` call `sanitizeNodeName(node.label)` directly with no dedup. Two nodes sharing a label therefore produce two identically-named `def`s and ambiguous transition strings — invalid Python. Please dedup consistently across function definitions, the action list, and transitions (or key generation off the stable `node.id` rather than the display label).
- **File:** `telemetry/ui/src/components/routes/graph-builder/utils/BurrCodeGenerator.ts`
  **Line:** ~106 (`getInputParameters`)
  **[SUGGESTION]** Input parameter names aren't deduplicated, so two input edges resolving to the same sanitized name would emit duplicate function parameters (a `SyntaxError`). Cheap to guard with the same dedup logic.
- **File:** `telemetry/ui/src/components/routes/graph-builder/utils/GraphExporter.ts`
  **Line:** ~57 (`metadata.created: new Date().toISOString()`)
  **[SUGGESTION]** This runs inside the `useMemo` recomputed on every node/edge change and is written to localStorage, so `created` resets to "now" on every edit. Set it once at creation (or omit from the persisted blob) — this also makes the generated JSON snapshot-testable.
- **File:** `telemetry/ui/src/components/routes/graph-builder/utils/ExampleLoader.ts`
  **Line:** ~52 (`validateExample`)
  **[SUGGESTION]** Great that this validator exists — please wire it into a unit test that runs over each bundled `ExampleGraph` and asserts it returns `[]`, so a malformed example can't ship silently.
- **File:** `telemetry/ui/src/components/routes/app/GraphView.tsx`
  **Line:** ~170 (removal of `getSmartEdge`, replaced by `getBezierPath`)
  **[NIT]** Dropping smart-edge routing is a minor visual regression for the observability graph (edges may overlap nodes). Fine if intentional given the dependency removal — just calling it out for confirmation.
- **File:** `telemetry/ui/src/components/nav/appcontainer.tsx`
  **Line:** ~394 (`<main>` layout rewrite)
  **[NIT]** This changes the shared layout used by every route (removing the previous `py-14 -my-1` margins for a flex column). Reasonable cleanup, but please smoke-test existing pages (projects, app view, annotations) for spacing regressions since it's outside the graph-builder scope.
- **File:** `telemetry/ui/src/components/routes/graph-builder/components/GraphBuilder.tsx`
  **Line:** ~230 (`useEffect(..., [])` initial-load hook with eslint-disable)
  **[NIT]** The mount-only load effect disables `exhaustive-deps`. Acceptable for a one-time restore, but a brief comment on why (avoid re-running on `initialGraph`/`loadGraphIntoCanvas` identity changes) would help future readers.

skrawcz:
Need to remember whether relationship with #687 means this is required or superseded

skrawcz:
need to rebase this against vite changes

Issue #213, opened 2024-05-26
Burr -- Google Sheets Persister


elijahbenizzy:
**Is your feature request related to a problem? Please describe.**
Heh, this is actually just a really cool idea. And, why not? Would make it super easy to debug, and google sheets is basically an extremely powerful (if not limited in scalability) database.

**Describe the solution you'd like**
Build a persister that saves to/loads from google sheets. 

- [ ] Determine data models -- partition key = one sheet? App = one tab? All in a drive?
- [ ] Determine parameters for data model
- [ ] Figure out best way to wire through secrets (ensuring that we don't accidentally track it...)
- [ ] Test
- [ ] Add CSV/Excel as well, based on this

See https://github.com/DAGWorks-Inc/burr/blob/26cfcd272a26c58e88c6ec84b18682b8ad87a9bd/burr/integrations/persisters/b_redis.py#L19. 

for a sample working persister

**Describe alternatives you've considered**
Not doing this for fear that someone will use it in prod (or maybe they should...).

cmm25:
I am requesting more information as I am interested in this issue

skrawcz:
@cmm25 sorry for the tardy response. What questions do you have here?

cmm25:
Seeing this as a good first issue I am interested in handling this but I cannot quite understand the needed requirements.

skrawcz:
@cmm25 sure let me try to unpack it.

1. The high level idea is that we could use google sheets as a database for persistence
2. e.g. use their python API to read and write data (see [their quickstart](https://developers.google.com/sheets/api/quickstart/python)) 

Things to implement:
1. A GoogleSheetsPersister (e.g. here's the [postgres one for example](https://github.com/DAGWorks-Inc/burr/blob/main/burr/integrations/persisters/postgresql.py))
2. The persister should implement the required methods and then subsequently use the python google sheets API to read/write appropriately.
3. Then an example application that uses it and exercises the functionality. 

How I would get started:
1. Have setup google credentials and are able to run https://developers.google.com/sheets/api/quickstart/python
2. Have forked the burr repository
3. Added the google python client as an optional dependency to the project
4.  created the new persister class
5. design the "data model" for the google sheet. It might be simplest to mirror the table structure from the postgresql persister?
6. check in with us for feedback / thoughts.

Does that help?

goutamk09:
@skrawcz  I'd be interested in working on this.

Proposed plan:

- Mirror the existing Postgres persister structure first (same logical schema).
[...]

PR #670, opened 2026-03-08
Add Google Sheets Persister (#213)

goutamk09:
Initial skeleton implementation for a GoogleSheetsPersister.

This follows the structure of existing Burr persisters and introduces a new persister that will store Burr state in Google Sheets using the Google Sheets Python API.

Planned next steps:
- Implement save/load logic
- Add example usage
- Add tests

### DIFF
--- a/test_sheets.py
+++ b/test_sheets.py
@@ -1,0 +1,15 @@
+from burr.integrations.persisters.b_google_sheets import GoogleSheetsPersister
+
+persister = GoogleSheetsPersister(
+    spreadsheet_id="1ur4GqM0tQQyCRn3lAj5bZk3DAeZBvrbYRwKok780hHc",
+    credentials_file="google_creds.json"
+)
+
+# Save some data
+persister.save("demo_app", "user1", {"hello": "world"})
+
+# Load the data
+data = persister.load("demo_app", "user1")
+
+print("Loaded data:", data)
+print("App IDs:", persister.list_app_ids())
--- a/.gitignore
+++ b/.gitignore
@@ -193,3 +193,4 @@ burr/tracking/server/build
 examples/*/statemachine
 examples/*/*/statemachine
 .vscode
+google_creds.json
--- a/burr/integrations/persisters/b_google_sheets.py
+++ b/burr/integrations/persisters/b_google_sheets.py
@@ -1,0 +1,29 @@
+import json
+from burr.core.persistence import BaseStatePersister
+
+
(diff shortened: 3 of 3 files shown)


goutamk09:
Current progress:

- [x] persister skeleton
- [x] basic save/load implementation
- [x] `list_app_ids()` implemented
- [x] verified the flow against a real Google Sheet

Next steps:
- add proper tests under `tests/`
- add a minimal example app
- align the interface more closely with existing persisters
- handle optional dependencies similar to other integrations

Happy to adjust the structure if there’s a preferred pattern for persisters in Burr.

test_sheets.py
```diff
@@ -0,0 +1,15 @@
+from burr.integrations.persisters.b_google_sheets import GoogleSheetsPersister
+
+persister = GoogleSheetsPersister(
+    spreadsheet_id="1ur4GqM0tQQyCRn3lAj5bZk3DAeZBvrbYRwKok780hHc",
+    credentials_file="google_creds.json"
+)
```
skrawcz: yeah how would we implement a test for this?

JiwaniZakir (commented):
The `save()` method always appends new rows, but `load()` returns the **first** matching row for a given `(app_id, partition_key)` pair — meaning after the first save, subsequent saves will never be loaded since the stale initial row is returned first. It should scan to the last matching row instead.

`list_app_ids()` iterates over `rows[1:]` while `load()` and `list()` iterate over `rows` — this inconsistency implies a header row is expected in some cases but not others. Either all methods should skip a header row or none should, and that expectation needs to be explicit (e.g., written during `__init__`).

`test_sheets.py` at the repo root contains a hardcoded real spreadsheet ID (`1ur4GqM0tQQyCRn3lAj5bZk3DAeZBvrbYRwKok780hHc`) and should not be committed; manual smoke tests like this belong either in a proper test file with mocks or removed entirely — the `.gitignore` addition for `google_creds.json` shows awareness of credential hygiene, but the spreadsheet ID itself is also worth protecting.

The method signatures for `save` and `load` don't appear to align with `BaseStatePersister`'s interface, which typically includes a `sequence_id` parameter to support ordered state history — without it, this persister won't interoperate correctly with Burr's state tracking machinery.

elijahbenizzy (commented):
This looks good, we should get this out and note it's under beta?

andreahlert:
@goutamk09 any updates on this?

PR #673, opened 2026-03-12
fix: wheel LICENSE should not reference examples-only MIT files

### LINKED ISSUE #641: license in apache_burr-0.41.0-py3-none-any.whl is incorrect
licenses/LICENSE is the source release license and includes details about MIT Licensed files in examples - these files do not appear in apache_burr-0.41.0-py3-none-any.whl

The LICENSE in convenience binaries like this whl should reflect what is the archive not be a random copy of the source license. The LICENSE in whl file should describe what is in the whl file.

andreahlert:
The wheel was shipping a LICENSE that mentioned MIT-licensed files from
examples/deep-researcher, but those files only exist in the source
distribution. Moved the third-party attribution to a dedicated
THIRD-PARTY-LICENSES file included only in the sdist.

Closes #641

## Changes

- Stripped the MIT section from LICENSE (now pure Apache 2.0)
- Added THIRD-PARTY-LICENSES with the MIT attribution for deep-researcher
- Included THIRD-PARTY-LICENSES in sdist via pyproject.toml (not in license-files, so the wheel stays clean)

## How I tested this

- Built wheel, confirmed LICENSE has no MIT section
- Built sdist, confirmed THIRD-PARTY-LICENSES is present
- test_release_config.py passing

## Notes

Per ASF policy, LICENSE must reflect what's actually in the distribution.

## Checklist

### DIFF
--- a/LICENSE-wheel
+++ b/LICENSE-wheel
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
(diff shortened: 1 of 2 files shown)


andreahlert:
@skrawcz @elijahbenizzy ready for review.

andreahlert:
@pjfanning Ah, let me see if I got this right now. The source tree LICENSE must keep the MIT attribution because `examples/deep-researcher` files are actually there. What needs to change is the wheel only, since it doesn't ship those examples and shouldn't carry the MIT notice.

I've reworked the approach: the source `LICENSE` is now untouched, and a separate `LICENSE-wheel` (Apache-only) is what Flit injects into the `.whl` via `license-files`. The sdist still gets the full `LICENSE` as before.

Can you take another look and see if this is what you had in mind?

PR #676, opened 2026-03-16
Add comparison operators to when() conditions

skrawcz:
## Summary

- Extends `Condition.when()` to support Django-style lookup operators via `__` suffixes
- Fully backward compatible — plain `when(key=value)` behavior is unchanged
- Adds 43 parametrized tests covering all operators, edge cases, and composition
- Updates transition documentation with operator reference

## Supported Operators

```python
from burr.core import when

when(age__gte=18)                          # age >= 18
when(age__gt=18)                           # age > 18
when(age__lt=18)                           # age < 18
when(age__lte=18)                          # age <= 18
when(age__ne=0)                            # age != 0
when(age__eq=18)                           # age == 18 (explicit)
when(status__in=["active", "pending"])     # membership
when(status__notin=["banned"])             # not in
when(tags__contains="python")              # collection contains value
```

Multiple kwargs are ANDed: `when(age__gte=18, status="active")`

Works with `~` (invert), `|` (or), and `&` (and) operators.

## Motivation

Previously, `when()` only supported exact equality. Users had to fall back to `expr()` string expressions for any comparison, losing type safety and IDE support. This change makes the most common comparison patterns expressible directly in `when()`.

## Test plan

- [x] All 12 existing condition tests pass unchanged (backward compat)
- [x] 25 parametrized operator tests (all operators × true/false cases)
- [x] 5 reads-extraction tests (single key, multiple keys, same key with two operators)
- [x] 8 name-formatting tests (symbol operators, word operators, plain equality)
- [x] Combined multi-operator test
- [x] Composition tests (~, |, &)
- [x] Invalid key validation test
- [x] Full test suite: 93 passed
     

## Checklist

### DIFF
--- a/burr/core/action.py
+++ b/burr/core/action.py
@@ -433,6 +433,8 @@ class Condition(Function):
         "in": ("in", lambda a, b: a in b),
         "notin": ("not in", lambda a, b: a not in b),
         "contains": ("contains", lambda a, b: b in a),
+        "is": ("is", lambda a, b: a is b),
+        "isnot": ("is not", lambda a, b: a is not b),
     }
 
     @classmethod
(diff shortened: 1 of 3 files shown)


elijahbenizzy (commented):
Looks good -- we should consider the python operators module to make this simpler

PR #680, opened 2026-03-18
fix: graceful stream shutdown on exceptions in streaming actions

### LINKED ISSUE #581: broken stream when handling raised exceptions
The streaming state machine when having a raised error inside one of the actions doesn't have a way of gracefully shutting down the stream in the "finally" clause of try/except/finally block.

This is the self-contained snipper to reproduce the behaviour. It's a FastAPI endpoint which you can run with python and then curl it. The code raises an error suddenly when the counter is equal 5, but you can comment it out to see that it should finish without this: `curl: (18) transfer closed with outstanding read data remaining`

Python sippet:

```
import datetime
from typing import AsyncGenerator
from fastapi import FastAPI
from fastapi.responses import StreamingResponse
import uvicorn

from burr.core import State, ApplicationBuilder, expr
from burr.core.action import streaming_action

app = FastAPI(title="Bug Report Demo")

@streaming_action(reads=["counter"], writes=["counter"])
async def increment(state: State):
    """Mock action for report - increment the counter by 1"""
    current_count = state["counter"]
    exception_occurred = False
    try:
        if current_count == 5:
            raise ValueError("Raising to show some exception from some function in the action!")
    except ValueError:
        yield {"response": "error! some helpful info for the frontend..."}, None
        exception_occurred = True
        raise
    finally:
        if exception_occurred:
            yield {"response": "gracefully stopping the stream..."}, None
            yield {}, state
    current_count += 1
    print("Count: ", current_count)
    yield {"response": "increment"}, None
    yield {}, state.update(counter=current_count)


@streaming_action(reads=["counter"], writes=[])
async def exit_counter(state: State):
    """Print the current count and the current time"""
    current_count = state["counter"]
    print(f"Finished counting to {current_count} at {datetime.datetime.now():%H:%M:%S %Y-%m-%d}")
    yield {"response": "exit_counter"}, None
    yield {}, state


async def build_app():
    """Build the state machine application"""
    return await (
        ApplicationBuilder()
        .with_actions(increment, exit_counter)
        .with_transitions(
            ("increment", "increment", expr("counter < 10")),
            ("increment", "exit_counter"),
        )
        .with_state(counter=0)
        .with_entrypoint("increment")
        .abuild()
    )

async def generate_stream() -> AsyncGenerator[str, None]:
    """Generate the stream of responses"""
    app = await build_app()
    async for action, streaming_container in app.astream_iterate(
        inputs={},
        halt_after=["exit_counter"],
    ):
        async for item in streaming_container:
            yield str(item) + "\n\n"

@app.get("/stream")
async def stream_endpoint():
    """Endpoint to demonstrate the bug report with streaming response"""
    return StreamingResponse(
        generate_stream(),
        media_type="text/plain",
    )

if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=8000)


```

## Stack Traces

Curl output: 

`curl -N http://localhost:8000/stream
{'response': 'increment'}

{'response': 'increment'}

{'response': 'increment'}

{'response': 'increment'}

{'response': 'increment'}

{'response': 'error! some helpful info for the frontend...'}

{'response': 'gracefully stopping the stream...'}

curl: (18) transfer closed with outstanding read data remaining`

# Expected behavior

I would expect the ability to gracefully close the stream i.e. stream back the error contents to the frontend (or any other consumer) after raising the Exception and then closing the burr operation by yielding the State instead of None. It seems now it's not really possible because the last yield in finally "gracefully stopping the stream..." is the last thing that is ever ran.

It would perhaps be good to be able to yield a textual response while closing the stream with:

`yield response_with_error_info, state`

andreahlert:
## Summary

- Wraps the generator consumption loop in `_run_single_step_streaming_action`, `_arun_single_step_streaming_action`, `_run_multi_step_streaming_action`, and `_arun_multi_step_streaming_action` with `try/except`
- When a streaming action catches an exception and yields a final state update (e.g. in a `finally` block), the stream now completes gracefully instead of propagating the exception and killing the connection
- If no state was yielded before the exception, the original behavior is preserved (exception propagates)

Closes #581

## Test plan

- [x] Reproduced the bug using the exact snippet from #581 in a Docker container (FastAPI + uvicorn). Confirmed `curl: (18) transfer closed with outstanding read data remaining` and `ValueError` stack trace on server
- [x] Applied fix and verified stream completes with exit code 0, server logs clean (no exceptions)
- [x] Ran all 29 streaming-related tests in `test_application.py` (all passing)
- [x] Ran full `test_application.py` + `test_action.py` suite: 175 tests, zero failures

### DIFF
--- a/burr/core/application.py
+++ b/burr/core/application.py
@@ -70,12 +70,7 @@ from burr.core.persistence import (
 from burr.core.state import State
 from burr.core.typing import ActionSchema, DictBasedTypingSystem, TypingSystem
 from burr.core.validation import BASE_ERROR_MESSAGE
-from burr.lifecycle.base import (
-    ExecuteMethod,
-    LifecycleAdapter,
-    PostRunStepHook,
-    PreRunStepHook,
-)
+from burr.lifecycle.base import ExecuteMethod, LifecycleAdapter, PostRunStepHook, PreRunStepHook
 from burr.lifecycle.internal import LifecycleAdapterSet
 from burr.visibility import tracing
 from burr.visibility.tracing import tracer_factory_context_var
@@ -94,9 +89,7 @@ StateType = TypeVar("StateType")
 StateTypeToSet = TypeVar("StateTypeToSet")
 
 
-def _validate_result(
-    result: Any, name: str, schema: ActionSchema = DEFAULT_SCHEMA
-) -> None:
+def _validate_result(result: Any, name: str, schema: ActionSchema = DEFAULT_SCHEMA) -> None:
     # TODO -- split out the action schema into input/output schema types
     # Currently they're tied together, but this doesn't make as much sense for single-step actions
     result_type = schema.intermediate_result_type()
@@ -165,9 +158,7 @@ def _remap_dunder_parameters(
     return inputs
 
 
-def _run_function(
-    function: Function, state: State, inputs: Dict[str, Any], name: str
-) -> dict:
+def _run_function(function: Function, state: State, inputs: Dict[str, Any], name: str) -> dict:
     """Runs a function, returning the result of running the function.
     Note this restricts the keys in the state to only those that the
     function reads.
@@ -187,9 +178,7 @@ def _run_function(
     function.validate_inputs(inputs)
     if "__context" in inputs or "__tracer" in inputs:
         # potentially need to remap the __context & __tracer variables
-        inputs = _remap_dunder_parameters(
-            function.run, inputs, ["__context", "__tracer"]
-        )
+        inputs = _remap_dunder_parameters(function.run, inputs, ["__context", "__tracer"])
     result = function.run(state_to_use, **inputs)
     _validate_result(result, name)
     return result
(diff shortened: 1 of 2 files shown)


andreahlert:
@skrawcz @kajocina This PR fixes the broken stream issue reported in #581. The generator consumption loops now catch exceptions gracefully when the action has already yielded a final state (e.g. in a `finally` block), instead of letting the exception kill the connection. All existing streaming tests pass. Would appreciate a review when you get a chance.

skrawcz:
pre-commit issue

andreahlert:
> Good fix for a real issue! The `try/except` approach in the single-step functions is clean — checking `state_update is None` correctly distinguishes between "generator crashed before completing" and "generator yielded a final state in its `finally` block."
> 
> Two items to address:
> 
> 1. **`caught_exc` should be logged** in the multi-step functions — it's currently assigned (`caught_exc = e`) but never read afterward. Please log it (e.g., `logger.warning(...)`) so that exceptions aren't silently swallowed. Users need visibility into what went wrong even when the stream completes gracefully.
> 2. **Tests needed.** The reproduction case from #581 translates directly into a test. Please add tests covering:
>    
>    * Streaming action with `try/except/finally` that yields state update in `finally` → completes gracefully
>    * Streaming action that raises without yielding state → exception still propagates
>    * Both single-step and multi-step variants
> 
> Also worth discussing: the multi-step guard (`if result is None: raise`) is more permissive than the single-step guard (`if state_update is None: raise`). In multi-step, if the generator crashes after yielding some items but before producing the "final" result, the reducer silently runs on partial data. Is this the intended behavior?

All three points addressed. Added logger.warning(...) with exc_info=True in all 4 functions (not just multi-step). Removed the dead caught_exc variable entirely. 

Added 8 tests covering graceful shutdown and exception propagation for both single-step and multi-step, sync and async. 

On the multi-step guard discussion: kept if result is None: raise but the warning now explicitly tells the user the reducer is running on potentially partial data, so there's full visibility without changing the current behavior.

Also ran additional review passes (on an isolated environment - thanks to Claude for that) that caught a few extra issues beyond what was flagged: added logging in the single-step functions too (not just multi-step), fixed a count increment asymmetry between sync and async single-step variants, and fixed test fixtures where stream_run wasn't a proper generator function (missing yield after raise).

IMHO, now we're ready to merge. But I think worth take an extra look.

skrawcz (approved):

PR #681, opened 2026-03-18
fix: support `async with` on async persister factory methods

### LINKED ISSUE #546: Async state persister doesn't work as a context manager
It seems that the AsyncSQLitePersister doesn't work in the expected way when trying to use it as a context manager.

# Current behavior

Trying to use AsyncSQLitePersister.from_values() with "async with" fails. Full example below.

## Stack Traces

(...)
    async with AsyncSQLitePersister.from_values(
TypeError: 'coroutine' object does not support the asynchronous context manager protocol

Python 3.11, MacOS

# Expected behavior

Expected the provided snippet to work as it follows how the context manager is shown in the tutorials during the application build step. The tutorial shows a regular sync versions, perhaps async ones need to be initialised differently and the tutorial updated?

import datetime
import asyncio

from burr.core import action, State, ApplicationBuilder, expr
from burr.integrations.persisters.b_aiosqlite import AsyncSQLitePersister

async def increment(state: State) -> State:
    """Increment the counter by 1"""

    # get the value from the `state`
    current_count = state["counter"]
    current_count += 1

    print("Count: ", current_count)

    # use `.update()` to create a new `State`
    return state.update(counter=current_count)

@action(reads=["counter"], writes=[])
async def exit_counter(state: State) -> State:
    """Print the current count and the current time"""
    current_count = state["counter"]

    print(f"Finished counting to {current_count} at {datetime.datetime.now():%H:%M:%S %Y-%m-%d}")

    return state

async def build_app():
    """This is where the problem shows up."""
    async with AsyncSQLitePersister.from_values(
            db_path=".burr.db",
            table_name="burr_state"
    ) as state_persister:
        app = await (
            ApplicationBuilder()
            .with_actions(increment, exit_counter)
            .with_transitions(
                ("increment", "increment", expr("counter < 10")),
                ("increment", "exit_counter"),
            )
            .with_state(counter=0)
            .with_entrypoint("increment")
            .with_state_persister(state_persister)
            .abuild()
        )
    return app

async def main():
    app = await build_app()
    action_obj, result, state = await app.arun(halt_after=["exit_counter"])

if __name__ == "__main__":
    asyncio.run(main())

andreahlert:
## Summary

Fixes #546

`AsyncSQLitePersister.from_values()` and `AsyncPostgreSQLPersister.from_values()` are async classmethods that return coroutines. Using them directly with `async with` fails with:

```
TypeError: 'coroutine' object does not support the asynchronous context manager protocol
```

This PR introduces `_AsyncPersisterContextManager`, a thin wrapper that implements both `__await__` and `__aenter__`/`__aexit__`, so factory methods now support both usage patterns:

```python
# await (backwards compatible)
persister = await AsyncSQLitePersister.from_values(db_path="test.db")

# async with (what the issue requested)
async with AsyncSQLitePersister.from_values(db_path="test.db") as persister:
    await persister.initialize()
    ...
```

The same fix is applied to `AsyncPostgreSQLPersister` in `b_asyncpg.py`.

## Test plan

- [x] Reproduced the original `TypeError` in a Docker container (Python 3.11)
- [x] Verified `async with ... from_values()` works after the fix
- [x] Verified `await ... from_values()` still works (backwards compatibility)
- [x] @skrawcz review
- [ ] @kajocina validate against original use case


andreahlert:
> Good solution to #546! The `_AsyncPersisterContextManager` approach is clean — supporting both `await` and `async with` from the same factory method is a nice UX improvement.
> 
> Two items to address:
> 
> 1. **Move `_AsyncPersisterContextManager` to `burr/common/`**. Having `b_asyncpg.py` import from `b_aiosqlite.py` is an unwanted cross-dependency between unrelated integrations. Something like `burr/common/types.py` or a new `burr/common/async_utils.py` would be a better home.
> 2. **Add tests for the `async with` pattern.** All existing tests use `await`, so the new context manager functionality is untested. At minimum, a test like:
>    ```python
>    async def test_async_sqlite_from_values_as_context_manager(tmp_path):
>        db_path = str(tmp_path / "test.db")
>        async with AsyncSQLitePersister.from_values(db_path=db_path) as persister:
>            await persister.initialize()
>            await persister.save("pk", "app1", 1, "pos", State({"k": "v"}), "completed")
>            loaded = await persister.load("pk", "app1")
>            assert loaded is not None
>    ```
> 
> Also checked: `AsyncRedisBasePersister.from_values` is already a regular (non-async) method, so it doesn't have this issue. The fix correctly covers the only two affected persisters.

Addressed all three points. While testing the fix in an isolated Docker container (Python 3.11), I caught two additional bugs:

- __aexit__ would crash with AttributeError if __aenter__ failed (e.g. bad db path), because self._persister was still None. Now guarded.
- The wrapper could be silently consumed twice (coroutine exhaustion). Added a _consumed flag that raises RuntimeError with a clear message.

Also fixed the parallelism.rst docs that were missing await on from_values() calls, and removed the now-redundant AsyncSQLiteContextManager helper from both test files.

LGTM. Could you please take an extra look?

PR #682, opened 2026-03-18
fix: pass partition_key to child PointerModel in fork/spawn relationships

### LINKED ISSUE #518: Bug with UI link to forked app with partition key
When an application is forked and uses a partition key, the link in the UI does not have that partition key in the URL.

# Current behavior
E.g. creating an app like this:
```python
forked_app = (
    ApplicationBuilder()
    .with_graph(base_graph) # this could be different...
    .initialize_from(
        tracker,
        resume_at_next_action=True,
        default_state={"chat_history": []},
        default_entrypoint="prompt",
        fork_from_app_id=app_id,
        fork_from_sequence_id=sequence_id,
        fork_from_partition_key=partition_key
    )
    .with_identifiers(partition_key=partition_key)
    .with_tracker(tracker, use_otel_tracing=True)
    .build()
)
```
Results in:

Image

Image

andreahlert:
## Changes

- Add `partition_key` parameter to `_log_child_relationships` so child `PointerModel` entries in `children.jsonl` have the correct `partition_key` instead of `None`
- Pass `partition_key` from `post_application_create` to `_log_child_relationships`
- Add regression test verifying `children.jsonl` contains the correct `partition_key` for forked apps

## How I tested this

- Added `test_fork_children_have_correct_partition_key` that creates a parent app with a partition key, forks from it, and asserts the child entry in `children.jsonl` has the correct `partition_key`
- All 16 existing tracking tests pass

## Notes

The bug only manifests when `partition_key` is explicitly used in forks. Without it, the `null` value accidentally matches the UI route, making it seem intermittent.

Closes #518

### DIFF
--- a/burr/tracking/client.py
+++ b/burr/tracking/client.py
@@ -207,6 +207,7 @@ class LocalTrackingClient(
         fork_parent_pointer_model: Optional[burr_types.ParentPointer],
         spawn_parent_pointer_model: Optional[burr_types.ParentPointer],
         app_id: str,
+        partition_key: Optional[str] = None,
     ):
         """Logs a child relationship. This is special as it does not log to the main log file. Rather
         it logs within the parent directory. Note this only exists to maintain (denormalized) bidirectional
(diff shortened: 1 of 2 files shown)


andreahlert:
@elijahbenizzy @skrawcz can you give this a spin and confirm the links are working on your end?

PR #683, opened 2026-03-18
core: add flexible_api decorator to fix mypy override errors

### LINKED ISSUE #457: Class based actions: Mypy not happy with defined parameters in the `run` method
For class-based actions, mypy is not happy with parameters to be defined in the method signature of the `run` method.

# Current behavior

I have copied the class-based action [example](https://burr.dagworks.io/concepts/actions/#class-based-actions) from doc:

```python
from burr.core import Action, State, ApplicationContext

class Counter(Action):
    @property
    def reads(self) -> list[str]:
        return ["counter"]

    def run(self,
            state: State,
            increment_by: int,
            # add __context here
            __context: ApplicationContext) -> dict:
        # can access the app_id from the __context
        print(__context.app_id, __context.partition_key, __context.sequence_id)
        return {"counter": state["counter"] + increment_by}

    @property
    def writes(self) -> list[str]:
        return ["counter"]

    def update(self, result: dict, state: State) -> State:
        return state.update(**result)

    @property
    def inputs(self) -> list[str]:
        # add __context here
        return ["increment_by", "__context"]
```

Then mypy is not happy with it with errors:

```bash
mypy src/demo_application.py
src/demo_application.py:8: error: Signature of "run" incompatible with supertype "Function"  [override]
src/demo_application.py:8: note:      Superclass:
src/demo_application.py:8: note:          def run(self, state: State[Any], **run_kwargs: Any) -> dict[Any, Any]
src/demo_application.py:8: note:      Subclass:
src/demo_application.py:8: note:          def run(self, state: State[Any], increment_by: int, ApplicationContext, /) -> dict[Any, Any]
Found 1 error in 1 file (checked 1 source file)
```

```bash
uv pip list | grep mypy
mypy               1.13.0
mypy-extensions    1.0.0

uv pip list | grep burr
burr               0.34.1

python --version
Python 3.12.5
```

andreahlert:
## Summary

- Adds a `flexible_api` decorator that can be applied to `run`, `stream_run`, or `run_and_update` overrides using explicit parameters instead of `**run_kwargs`
- Prevents mypy `[override]` errors caused by narrowing the base-class signature
- Exported from `burr.core` for easy access

## Context

When users define class-based actions with typed parameters in `run()` (which is the documented pattern), mypy reports an `[override]` error because the subclass signature is incompatible with `Function.run(self, state, **run_kwargs)`.

The decorator wraps the method so mypy sees `Callable[..., Any]`, which is compatible with the base signature.

### Before (triggers mypy error)
```python
class Counter(Action):
    def run(self, state: State, increment_by: int) -> dict:  # mypy: [override]
        return {"counter": state["counter"] + increment_by}
```

### After (no mypy error)
```python
from burr.core import flexible_api

class Counter(Action):
    @flexible_api
    def run(self, state: State, increment_by: int) -> dict:  # ok
        return {"counter": state["counter"] + increment_by}
```

## Test plan

- Tested in isolated Docker environment (Python 3.12, mypy 1.19.1) confirming the decorator eliminates the `[override]` error
- Class without decorator still triggers the error as expected (no behavior change for existing code)

Closes #457

### DIFF
--- a/burr/core/__init__.py
+++ b/burr/core/__init__.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from burr.core.action import Action, Condition, Result, action, default, expr, flexible_api, when
+from burr.core.action import Action, Condition, Result, action, default, expr, type_eraser, when
 from burr.core.application import (
     Application,
     ApplicationBuilder,
@@ -35,7 +35,7 @@ __all__ = [
     "Condition",
     "default",
     "expr",
-    "flexible_api",
+    "type_eraser",
     "Result",
     "State",
     "when",
--- a/burr/core/__init__.py
+++ b/burr/core/__init__.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from burr.core.action import Action, Condition, Result, action, default, expr, when
+from burr.core.action import Action, Condition, Result, action, default, expr, flexible_api, when
 from burr.core.application import (
     Application,
     ApplicationBuilder,
@@ -35,6 +35,7 @@ __all__ = [
     "Condition",
     "default",
     "expr",
+    "flexible_api",
     "Result",
     "State",
     "when",
--- a/burr/core/action.py
+++ b/burr/core/action.py
@@ -143,6 +143,7 @@ def flexible_api(func: Callable[..., Any]) -> Callable[..., Any]:
 
     return wrapper
 
+
 # This is here to make accessing the pydantic actions easier
 # we just attach them to action so you can call `@action.pyddantic...`
 # The IDE will like it better and thus be able to auto-complete/type-check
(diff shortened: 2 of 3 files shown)


skrawcz (changes requested):
let's rethink the approach.

skrawcz:
Okay @andreahlert after digging in a bit more. Let's name the decorator `@type_eraser` or something -- since that will be clearer what the purpose is.

andreahlert:
> Okay @andreahlert after digging in a bit more. Let's name the decorator `@type_eraser` or something -- since that will be clearer what the purpose is.

Nice. Renamed to @type_eraser as suggested and  found and fixed an extra bug where the original @wraps wrapper was breaking is_async() detection for async and generator functions. 

Added some extra tests covering sync and etc. 

Ready to review

skrawcz (approved):

PR #684, opened 2026-03-18
fix: instantiate instrumentor in init_instruments()

### LINKED ISSUE #408: bug: `burr.integrations.opentelemetry.init_instruments()` doesn't work consistently.
This requires some investigation as it currently fails silently. We can tell it fails because no telemetry appears in Burr UI. In case of success, the two should lead to equivalent Burr UI telemetry displayed:

```python
from opentelemetry.instrumentation.openai import OpenAIInstrumentor

OpenAIInstrumentor().instrument()

# build and run the app
app = ApplicationBuilder().with_tracker(..., use_otel_tracing=True).build()
app.run(...)
```

```python
from burr.integrations.opentelemetry import init_instruments

init_instruments("openai")

# build and run the app
app = ApplicationBuilder().with_tracker(..., use_otel_tracing=True).build()
app.run(...)
```

Sidenote: the benefit of the Burr integration is that it allows to instrument multiple libraries with the same import by changing the strings passed as arg

andreahlert:
## Summary
- `_init_instrument` was calling `.is_instrumented_by_opentelemetry` and `.instrument()` on the **class** instead of an **instance**
- The `property` descriptor on the class is always truthy, so instrumentation was silently skipped with a `"is already instrumented"` debug log
- Even if it reached `.instrument()`, the call would fail with `TypeError: missing 1 required positional argument: 'self'`, caught and swallowed by the bare `except BaseException`

Fixes the class reference to instantiate it first, matching the standard OpenTelemetry instrumentor API.

Closes #408

## Test plan
- [x] Reproduced in a clean Docker container (Python 3.11, burr 0.41.0, opentelemetry-instrumentation-openai) confirming `is_instrumented_by_opentelemetry = False` after `init_instruments("openai")`
- [x] Applied fix and verified `is_instrumented_by_opentelemetry = True` after `init_instruments("openai")`

### DIFF
--- a/burr/integrations/opentelemetry.py
+++ b/burr/integrations/opentelemetry.py
@@ -665,7 +665,8 @@ def _init_instrument(
 
     try:
         instrumentation_module = importlib.import_module(instrumentation_module_name)
-        instrumentor = getattr(instrumentation_module, instrumentor_name)
+        instrumentor_cls = getattr(instrumentation_module, instrumentor_name)
+        instrumentor = instrumentor_cls()
         if instrumentor.is_instrumented_by_opentelemetry:
             logger.debug(f"`{module_name}` is already instrumented.")
         else:


andreahlert:
@elijahbenizzy @skrawcz @zilto fix submitted, tested locally and LGTM. Ready for review.

PR #687, opened 2026-03-23
feat: add visual builder + workspace mode to Burr UI

andreahlert:
Building on #667 (Graph Builder by @skrawcz), this adds a workspace mode and significantly enhanced builder that turns Burr's UI into a development environment.

## What's new

**Visual Builder (enhanced)**
- 9 node types: Action, Input, Result, LLM Call, API Call, Code, Streaming, Loop, Router
- Each type has specialized property editors (LLM: provider/model, API: url/method, Code: inline editor, Router: branch conditions)
- Inline + button on edges (Activepieces-style) with searchable type picker
- Recursive layout algorithm (replaces dagre) supporting loop arcs and router fan-out/fan-in
- Node type picker with search bar and categories (I/O, Logic, Integrations)

**Monaco Editor (split view)**
- Builder canvas on the left, VS Code-style editor on the right
- Multi-file code generation: actions.py, app.py, run.py, requirements.txt
- File explorer sidebar with tabs (open/close files)
- Bidirectional sync: edit code and the graph updates, edit graph and the code updates
- Full Ctrl+C/V/Z support in both canvas and editor

**Project save/load**
- File menu (New, Save, Open, Download .py)
- Projects persisted server-side in ~/.burr/builder_projects/
- Ctrl+S to save

**Workspace integration**
- Backend workspace router with file browsing, code viewing, and script execution (SSE streaming)
- Link workspace directories to projects
- Path traversal protection with workspace validation

**Other improvements**
- Undo/redo with 50-state history
- Keyboard shortcuts (Delete to remove nodes, Ctrl+C/V to copy/paste nodes)
- Code generation includes proper imports, sanitized identifiers, and with_tracker()

## Screenshots

**Builder with nodes and generated code**
02-builder-with-nodes

**Node editor with LLM Call configuration**
03-builder-node-editor

**Searchable node type picker**
04-builder-picker

## Technical details

**Backend** (Python)
- `burr/tracking/server/workspace.py`: new APIRouter with endpoints for file tree, file content, script execution (SSE), process management, workspace links, and builder project persistence
- Modified `run.py` to mount workspace router, `schema.py` to add `supports_workspace`

**Frontend** (~6,700 new lines)
- `utils/`: builderTypes, codeGenerator (9 node types + multi-file), codeParser (bidirectional sync), flowLayout (recursive layout), treeUtils (immutable tree ops), stateFlow
- `hooks/useBuilderState.ts`: tree-based state with undo/redo, keyboard shortcuts, save/load
- `builder/`: BuilderView (split layout), BuilderGraph (ReactFlow with custom node/edge types), NodeEditor, NodeTypePicker (portal-based), 4 node components, 5 edge components with inline + buttons
- `workspace/`: FileExplorer, CodeViewer, RunTerminal, WorkspaceSelector
- `project/`: ProjectWorkspaceView, ActivityBar, SidePanel, TabBar, TrackingSidePanel

## Dependencies added
- `@monaco-editor/react` (MIT)
- `@tanstack/react-query` (MIT)
- `@xyflow/react` (MIT)
- `react-joyride` (MIT)

All MIT license, ASF Category A compatible.

## How I tested
- `npx tsc --noEmit` zero errors
- Backend endpoints tested with curl
- Manual E2E: create flow, add nodes via toolbar and inline +, edit properties, see code update, save/load projects

## Checklist
- [x] Code passed the pre-commit check
- [x] New functions are documented


skrawcz:
@andreahlert I can't get this to run locally -- did you commit everything? otherwise see the CI failures.

skrawcz:
Also need some example to load to help someone get started.

andreahlert:
Thanks for reviewing. Didn't have the time to check it properly, I'll do it as soon as possible. Target: 03-30-2026

andreahlert:
Thanks for trying it out. Two things:

1. **Missing pieces**: This PR is meant to be applied on top of #667 (your graph builder). On main alone, the builder route works but some of the other UI improvements (HomeView, SearchView, etc.) won't be there. The builder itself at `/builder` should load fine since all its deps are self-contained. I'll add a note about this in the PR description.

2. **Example graphs**: Good call. I'll add a few pre-loaded examples (chatbot flow, data pipeline, agent with tools) that load from the File > Open menu so someone can see the builder in action without building from scratch.

Working on both now.

JiwaniZakir (commented):
In `run.py`, the shutdown sequence in `lifespan` calls `await workspace.cleanup_processes()` directly before `await backend.lifespan(app).__anext__()`, but there's no `try/finally` wrapping these calls. If `cleanup_processes()` raises an exception, the backend's own cleanup will be silently skipped, potentially leaking database connections or other resources. The backend shutdown should be guarded with a `try/finally` block regardless of what happens in workspace cleanup.

Additionally, `supports_workspace=True` is hardcoded unconditionally in `_get_app_spec()` — this means the UI will always advertise workspace support even if the feature is disabled or broken at runtime. This should be derived from a config flag or at least check that the workspace module initialized successfully, similar to how `is_snapshotting_backend` and `is_annotations_backend` are computed from the actual backend capabilities.

The `_validate_workspace` allowlist approach in `workspace.py` is a reasonable security boundary, but `_read_links()` is called on every file-browsing request with no caching. If the links file is read from disk on each API call under load, this could become a bottleneck — a short-lived in-process cache or memoization with invalidation on write would be worth considering.

skrawcz:
Screenshot 2026-03-29 at 09 25 22
Screenshot 2026-03-29 at 09 25 26

validated things load locally at least.

elijahbenizzy (commented):
Played around some -- looking great! Node types should probably have specific implementations but I think this is a good start to put out there.

andreahlert:
All review feedback addressed. Rebased clean on main, no conflicts.

@JiwaniZakir your three points: `cleanup_processes()` is wrapped in try/finally now so backend shutdown always runs, `supports_workspace` is derived from `workspace.is_available()` instead of hardcoded, and `_read_links()` has an in-process cache with invalidation on write.

@skrawcz docstrings on every function/class in workspace.py, including the inner ones. Examples were already in from the graph-builder merge.

@elijahbenizzy opened #735 to track the node type implementations as a follow-up. Agreed it's fine for the initial merge.

Ready for re-review whenever you get a chance.

burr/tracking/server/workspace.py
```diff
@@ -0,0 +1,700 @@
...
+        return True
+
+
+# --- Workspace Links ---
+
+_LINKS_PATH = os.path.join(os.path.expanduser("~/.burr"), "workspace_links.json")
+
+# In-process cache for workspace links to avoid disk reads on every request.
+_links_cache: Optional[dict] = None
+_links_cache_lock = threading.Lock()
```
skrawcz: we don't need this. we run fastapi in an asyncio context.

burr/tracking/server/workspace.py
```diff
@@ -0,0 +1,700 @@
...
+        if _links_cache is not None:
+            return _links_cache
+    if os.path.exists(_LINKS_PATH):
+        with open(_LINKS_PATH, "r") as f:
+            data = json_module.load(f)
+    else:
+        data = {}
+    with _links_cache_lock:
+        _links_cache = data
+    return data
```
skrawcz, line 253: due to the way python runs, this should be atomic in an asyncio context.

burr/tracking/server/workspace.py
```diff
@@ -0,0 +1,700 @@
...
+    """Write workspace-to-project links to ~/.burr/workspace_links.json.
+
+    Invalidates the in-process cache so subsequent reads pick up changes.
+    """
+    global _links_cache
+    os.makedirs(os.path.dirname(_LINKS_PATH), exist_ok=True)
+    with open(_LINKS_PATH, "w") as f:
+        json_module.dump(data, f, indent=2)
+    with _links_cache_lock:
+        _links_cache = data
```
skrawcz: again we don't need the lock.

skrawcz (commented):
just one nit around threading.lock

burr/tracking/server/workspace.py
andreahlert: yeah fair, my bad. threw the lock in out of habit (check-then-act on a global), but ofc it's all async + no awaits so it's already serialized. ripping out the lock + threading import.

skrawcz:
we just need to rebase -- and run locally and double check deps added

PR #741, opened 2026-04-18
ci: improve artifact verification flow

Haricharanpanjwani:
Closes #720.

Improve the Apache release artifact verification flow used by voters and release managers.

## Summary

- add structured pass/fail/skip result collection and summary rendering in `scripts/verify_apache_artifacts.py`
- verify `LICENSE`, `NOTICE`, and `DISCLAIMER` contents in release artifacts
- verify wheel-specific license files, including `LICENSE-wheel`
- add reproducible rebuild verification by rebuilding from the source artifact and comparing outputs against release artifacts
- integrate Apache RAT results into the verifier summary flow
- support RAT 0.18 output quirks, including log-line preambles, trailing summary lines, and relative-target invocation from extracted temp directories
- fix `.rat-excludes` patterns so RAT 0.18 applies the intended exclusions during end-to-end verification
- align release-wheel and rebuild-wheel packaging so reproducible comparison passes against RC-style artifacts
- add optional vote email generation via `--vote-email` and `--vote-email-output`
- add tests covering metadata verification, wheel license checks, rebuild comparison flow, vote email rendering, and RAT parser/invocation compatibility
- address review feedback by converting the verifier tests to `pytest`, making wheel metadata requirements explicit, removing the generic `build_tool` path, and isolating rebuild environment changes to subprocess execution

## How I tested this

- ran `python -m pytest tests/test_verify_apache_artifacts.py`
- ran `PYTHONPYCACHEPREFIX=/tmp/pycache python -m py_compile scripts/apache_release.py scripts/verify_apache_artifacts.py tests/test_verify_apache_artifacts.py`
- ran `python scripts/verify_apache_artifacts.py --help`
- ran `python scripts/verify_apache_artifacts.py all --help`
- created a clean worktree from the PR commit and set up a fresh `.venv`
- built RC-style artifacts with `python scripts/apache_release.py archive 0.41.0 0`, `python scripts/apache_release.py sdist 0.41.0 0`, and `python scripts/apache_release.py wheel 0.41.0 0`
- ran `python scripts/verify_apache_artifacts.py signatures --artifacts-dir dist`
- ran `python scripts/verify_apache_artifacts.py artifacts --artifacts-dir dist`
- ran `python scripts/verify_apache_artifacts.py licenses --artifacts-dir dist --rat-jar /Users/hpanjwani/Downloads/apache-rat-0.18/apache-rat-0.18.jar`
- ran `python scripts/verify_apache_artifacts.py reproducible --artifacts-dir dist`
- ran `python scripts/verify_apache_artifacts.py all --artifacts-dir dist --rat-jar /Users/hpanjwani/Downloads/apache-rat-0.18/apache-rat-0.18.jar --vote-email --vote-email-output /tmp/burr-vote-email.txt`

## Notes

- this PR is intentionally scoped to issue #720 and covers the requested acceptance criteria: reproducible rebuild verification, RAT integration, artifact metadata checks, clear pass/fail summaries, and optional vote email generation
- end-to-end verification was done against the RC-style artifact set produced by `scripts/apache_release.py`
- the main reviewer-facing areas are `scripts/verify_apache_artifacts.py`, `tests/test_verify_apache_artifacts.py`, `.rat-excludes`, and the release/rebuild alignment in `scripts/apache_release.py`

## Checklist

- [x] Code passed local verification for this change
- [x] Project documentation/help text was updated where applicable


Haricharanpanjwani:
I ran the new verifier flow locally in a clean clone to exercise the issue #720 implementation paths.

What the local run confirmed:
- `LICENSE` / `NOTICE` / `DISCLAIMER` checks pass on built artifacts
- wheel license-file checks pass, including `LICENSE-wheel`
- structured pass/fail summary output is working
- vote email draft generation is working
- the reproducible rebuild verification path is active: it rebuilt from source and compared rebuilt outputs against the artifacts under test

Caveats from the local run:
- the local artifact set was unsigned, so `.asc` and `.sha512` checks failed as expected
- Apache RAT was not exercised in that run unless a local `--rat-jar` was provided
- the reproducible checksum comparison failed against locally built artifacts, which shows the check is active rather than silently passing

Reproducibility-specific caveats encountered:
- local `flit build` artifacts do not necessarily match the Apache release-script naming/layout without additional release-style handling
- a manual rename was needed to exercise the source-artifact path in the verifier outside the full release flow
- a passing reproducible checksum comparison depends on reproducible release artifact generation, including the work tracked in #716
- in other words, this local run validates the new verification logic, but a fully green reproducibility result still depends on the release artifacts themselves being reproducible

So this comment is not claiming a fully green release verification from local demo artifacts. It is documenting that the new verifier checks are implemented, executable, and correctly detect both success and failure conditions.

Haricharanpanjwani:
Follow-up update pushed in `9c8417dc` (`ci: support modern Apache RAT output`).

This came out of local validation with Apache RAT 0.18:
- RAT emits INFO/WARN lines before the XML payload
- RAT 0.18 uses a newer `<license ...>` XML shape than the parser originally expected
- rerunning the verifier could also pick up generated `rat-report-*` files as artifacts

This follow-up makes the verifier more robust and backward compatible by:
- stripping log preamble lines before XML parsing
- supporting both the older RAT XML form and the RAT 0.18 license element shape
- ignoring generated `rat-report-*` files during artifact enumeration
- adding regression tests for those cases

tests/test_verify_apache_artifacts.py
```diff
@@ -0,0 +1,268 @@
...
+            temp_path.unlink()
+
+
+def _write_wheel(path: Path, files: dict[str, bytes]) -> None:
+    with zipfile.ZipFile(path, "w") as wheel:
+        for name, content in files.items():
+            wheel.writestr(name, content)
+
+
+class VerifyApacheArtifactsTests(unittest.TestCase):
```
skrawcz: can we change to pytest please.

skrawcz (commented):
Nice improvement to the release verification flow. The structured `VerificationSummary` / `CheckResult` approach is clean and makes the output much more useful for voters. The RAT 0.18 compatibility work (log-line parsing, relative-target invocation, new XML format handling) and reproducible build verification are both valuable additions.

One question on `_build_reproducible_wheel` (line 772): the `os.environ.clear()` followed by `os.environ.update(original_env)` pattern is risky if anything goes wrong between those two calls — the process would be left with an empty environment. Could you pass the modified env through to the subprocess calls instead of mutating `os.environ` directly?

A few smaller suggestions inline. All 9 tests pass locally and the overall structure is solid.

scripts/verify_apache_artifacts.py
```diff
@@ -443,59 +684,295 @@ def verify_licenses(artifacts_dir: str, rat_jar_path: str, report_only: bool = F
...
+        return False, f"wheel build exited with code {exc.code}"
+    except Exception as exc:
+        return False, str(exc)
+    finally:
+        os.chdir(original_cwd)
+        os.environ.clear()
+        os.environ.update(original_env)
+
+
+def _build_reproducible_artifacts(source_artifact: str, build_tool: list[str], output_dir: str) -> tuple[bool, str]:
```
skrawcz: _build_reproducible_artifacts accepts a build_tool parameter but the sdist build on line 793 hardcodes flit. Since flit is the required build tool for Burr, please remove the build_tool parameter here and in _detect_build_command — just use flit directly and fail if it's not available.

scripts/verify_apache_artifacts.py
```diff
@@ -87,354 +237,445 @@ def _verify_artifact_signature(artifact_path: str, signature_path: str) -> bool:
...
+                expected_files,
+                summary,
+            ):
+                all_valid = False
+        elif artifact_name.endswith(".whl"):
+            file_bytes = _wheel_file_bytes(artifact_path)
+            if not _verify_required_text_files(
+                artifact_name,
+                file_bytes,
+                REQUIRED_TEXT_FILES[1:] + WHEEL_LICENSE_FILES,
```
skrawcz: Nit: REQUIRED_TEXT_FILES[1:] silently skips LICENSE for wheels. A named constant (e.g., WHEEL_REQUIRED_FILES = ("NOTICE", "DISCLAIMER") + WHEEL_LICENSE_FILES) would make the intent clearer.

Haricharanpanjwani: Good call. I replaced the slice with an explicit `WHEEL_REQUIRED_TEXT_FILES` constant so the wheel-specific requirements are named directly instead of being implied by tuple slicing.

scripts/verify_apache_artifacts.py
Haricharanpanjwani: Agreed. I removed the generic `build_tool` path and `_detect_build_command` from the reproducible rebuild flow. The verifier now assumes `flit` directly, which matches Burr’s actual release build process and keeps the rebuild path simpler and more explicit.

tests/test_verify_apache_artifacts.py
Haricharanpanjwani: Done. I converted `tests/test_verify_apache_artifacts.py` to pytest style and removed the `unittest` scaffolding. The coverage is the same, but it now matches the repo’s preferred test style better.

scripts/verify_apache_artifacts.py
Haricharanpanjwani: Agreed. I removed the parent-process environment mutation and switched the reproducible wheel build to run  in a subprocess with an explicit environment. That keeps the behavior scoped to the build invocation and avoids leaving the process in a bad state if something fails mid-restore.

andreahlert:
tks @Haricharanpanjwani !

looked at this. the CI failure is just a missing init.py, easy fix one thing tho, get_first_matching_hook means only one interceptor per action right? might want to document that EOF

elijahbenizzy:
MInd rebasing? Looks reasonable. I added some changes so can you rebase? Then I'll look again.

Haricharanpanjwani:
> tks @Haricharanpanjwani !
> 
> looked at this. the CI failure is just a missing init.py, easy fix one thing tho, get_first_matching_hook means only one interceptor per action right? might want to document that EOF

Rebased onto the latest `main` and force-pushed. The  branch now includes the missing `tests/__init__.py`  fix for the CI failure as well.

andreahlert (approved):
LGTM

just  `_render_template` still has a lazy `from jinja2 import ...` inside the function body — @skrawcz  asked to just make it required, so might as well move it to the top-level imports

Haricharanpanjwani:
> just `_render_template` still has a lazy `from jinja2 import ...` inside the function body — @skrawcz asked to just make it required, so might as well move it to the top-level imports

@andreahlert I checked the current branch tip after the rebase and that `_render_template` helper / lazy `jinja2` import is no longer present in the PR changes, so there isn’t anything left to move to top-level imports here.

Haricharanpanjwani:
Re-surfacing this for final review/merge.

The rebase request from April 25, 2026 was addressed on April 28, 2026, and the current PR state is approved + mergeable. The branch also includes the follow-up fix for the earlier CI issue (`tests/__init__.py`).

If there is any remaining blocker on `#720`, please call it out explicitly and I can address it. Otherwise this looks ready for a final maintainer pass.

@elijahbenizzy @andreahlert

PR #675, opened 2026-03-12
feat: implement counter app demo in Burr UI

andreahlert:
## Summary

- Add FastAPI server for the hello-world-counter example with `/count`, `/state` and `/create` endpoints, following the multi-modal-chatbot server pattern
- Register counter router in the main Burr tracking server (`run.py`)
- Implement Counter React component with increment button and live telemetry view, replacing the WIP placeholder
- Add `CounterState` model and API service methods to the frontend

Closes #69

## Test plan

- [x] Python: server module imports, counter increments correctly, Pydantic model validates
- [x] TypeScript: `tsc --noEmit` passes with no errors
- [x] End-to-end: all endpoints tested via curl against a running Burr server (`POST /create`, `GET /state`, `POST /count`)
- [x] Tracking integration: steps are recorded and visible via `/apps` and application logs endpoints
- [x] OpenAPI spec: counter endpoints appear correctly in the generated schema

PR #745, opened 2026-04-20
fix: address RC2 vote feedback

elijahbenizzy:
## Summary

Addresses feedback from Stefan Krawczyk on the 0.42.0-incubating RC2 vote:

- Add `burr/tracking/server/__init__.py` — makes it a proper Python package so `importlib.resources` can find static files when installed from wheel
- Add `tests/__init__.py` — allows `pytest tests/` to work
- Add ASF header to `examples/deployment/aws/terraform/tutorial.md`
- Update `.rat-excludes`: fix patterns for `burr/examples/` symlink duplicates, add `.github/` templates and image file extensions

## Test plan
- [ ] Build wheel, install in fresh venv, run `burr` outside source tree — server should start
- [ ] `pytest tests/` should discover and run tests
- [ ] Run Apache RAT — zero unapproved files

PR #746, opened 2026-04-21
ci: release validation workflow (smoke install + RAT)

elijahbenizzy:
## Summary

Adds a `release-validation` GitHub Actions workflow that builds the full release pipeline on every PR and catches the class of bugs that have broken recent RCs (RC1 license/header, RC2 missing `hello-world-counter`).

The workflow:

1. **build-artifacts** — runs `scripts/apache_release.py all --skip-signing --no-upload` to produce the git archive, sdist, and wheel; runs Apache RAT on the tarballs. Any unapproved license or missing required file fails the job.
2. **install-and-smoke** — on Python 3.10/3.11/3.12, downloads the built wheel, installs it into a fresh venv outside the source tree, and runs `scripts/ci_smoke_server.py`: imports `burr.tracking.server.run` (catches missing-example bugs), starts the server on a free port, runs a tracked app, asserts the server sees the project.

Only `build-artifacts` and `install-and-smoke (3.12)` are required in `.asf.yaml`; 3.10 and 3.11 are informational.

## Commits

1. `--skip-signing` flag in `apache_release.py` + new `scripts/ci_smoke_server.py` helper
2. Workflow + `.asf.yaml` required-status-checks update

## Test plan

- [ ] `build-artifacts` is green on this PR
- [ ] `install-and-smoke (3.12)` is green on this PR
- [ ] Open a follow-up PR that e.g. removes a required example from the release config; confirm `install-and-smoke` fails (would have caught PR #675's regression)

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -63,7 +63,9 @@ github:
         # strict means "Require branches to be up to date before merging".
         strict: false
         # contexts are the names of checks that must pass
-        # contexts:
+        contexts:
+          - "Release Validation / build-artifacts"
+          - "Release Validation / install-and-smoke (3.12)"
       required_pull_request_reviews:
         dismiss_stale_reviews: false
         require_code_owner_reviews: false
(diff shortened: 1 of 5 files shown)


elijahbenizzy:
Consolidated into #745 — same fixes plus the workflow + RAT pattern fix.

PR #749, opened 2026-05-02
ci: add RC promotion command

Haricharanpanjwani:
Part of #715.
Closes #717.

## Summary

- add a new `promote` subcommand to `scripts/apache_release.py` to promote a voted RC from Apache SVN `dist/dev` to `dist/release`
- validate the expected source archive, sdist, wheel, and matching `.asc` / `.sha512` files before promotion
- remove the current release artifacts before copying in the promoted artifacts, matching ASF "latest only" release policy
- preserve `KEYS` in `dist/release` during promotion so signature verification material is not deleted
- print the final `twine upload` command for the promoted release artifacts
- support `--dry-run` so the command shows the intended release cleanup, copy, and commit steps without committing
- support incubating -> TLP path transitions through `--dev-svn-root` and `--release-svn-root` overrides
- document the new promotion flow in `scripts/README.md`
- add focused pytest coverage for RC parsing, artifact validation, release cleanup behavior, rename behavior, PyPI command rendering, and dry-run orchestration

## How I tested this

- ran `python3 -m py_compile scripts/apache_release.py tests/test_apache_release.py`
- ran a direct Python smoke check for RC parsing and release cleanup behavior, including preserving `KEYS`
- ran `python3 scripts/apache_release.py promote --help`

## Notes

- the promotion flow is implemented in `scripts/apache_release.py` rather than as a separate script so it stays alongside the existing release build, upload, and verification commands
- the command defaults to incubator SVN paths today, but accepts `--dev-svn-root` and `--release-svn-root` so it can handle the post-incubation TLP path change without code changes
- this PR covers the issue acceptance criteria by adding a single promotion command, validating the required artifacts before promotion, supporting incubator/TLP SVN root overrides, and providing a non-committing dry-run mode
- the release cleanup now explicitly preserves `KEYS`, which lives in the same `dist/release` directory and must remain available for signature verification
- I did not run a full end-to-end `promote --dry-run` against live SVN in this environment; the command still requires SVN access to validate the real RC and release paths before planning the promotion steps

## Checklist

- [x] Project documentation has been updated for the new command

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -91,6 +91,11 @@ python scripts/apache_release.py wheel 0.41.0 0        # Wheel dist
 python scripts/apache_release.py upload 0.41.0 0 your_apache_id
 python scripts/apache_release.py upload 0.41.0 0 your_apache_id --dry-run  # Test first
 
+# Promote a voted RC from dist/dev to dist/release
+python scripts/apache_release.py promote 0.41.0-RC0 your_apache_id
+python scripts/apache_release.py promote 0.41.0-RC0 your_apache_id --dry-run
+python scripts/apache_release.py promote 0.41.0-RC0 your_apache_id --release-svn-root https://dist.apache.org/repos/dist/release/burr  # TLP path override
+
 # Verify artifacts locally
 python scripts/apache_release.py verify 0.41.0 0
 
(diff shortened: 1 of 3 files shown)


Haricharanpanjwani:
@skrawcz @andreahlert can you review this PR?

andreahlert:
@Haricharanpanjwani can you check the CI?

Haricharanpanjwani:
> @Haricharanpanjwani can you check the CI?

@andreahlert Fixed the formatting. pre-commit should pass now

andreahlert:
Thanks @Haricharanpanjwani! I pushed a couple of commits to get this green and to correct the release-side layout.

**Formatting**
- Ran `black` on `scripts/apache_release.py` and `tests/test_apache_release.py` (the `pre-commit` black hook was failing).

**Release layout fix**
The original flow copied the artifacts flat into the `dist/release` project root and `svn rm`'d every existing entry except `KEYS`. But our `dist/release/incubator/burr` publishes each release under a per-version subdirectory (`0.41.0/`, `0.42.0/`, …) and keeps prior releases alongside the shared `KEYS`. Run as-is, `promote` would have placed the artifacts in the wrong location and deleted the already-published `0.41.0` and `0.42.0` releases.

Reworked `promote` to a single atomic server-side copy:

```
svn cp <dev_rc_url> <release_root>/<version> -m "..."
```

- additive: existing release directories and `KEYS` are untouched by construction
- no local download of the release tree
- refuses to run if the target version directory already exists
- the RC is still checked out locally only to validate the expected artifacts and their `.asc` / `.sha512` companions before promotion

This also let me drop the flat-copy/remove/commit helpers and the unused RC-suffix rename (the release artifact names never carry the RC number, it lives in the dev directory name). README and tests updated to match; net `-78` lines.

One caveat: the SVN workflow is unit-tested with mocked `svn`, but I have not verified it end-to-end against live ASF dist.

**Minor follow-up (non-blocking):** `_promotion_source_url` and the artifact patterns hardcode `-incubating`, so the `--dev-svn-root` / `--release-svn-root` overrides won't fully cover the post-graduation TLP paths yet. Fine while we're incubating; worth deriving the `-incubating` segment from context in a later change.

andreahlert (approved):
LGTM. Just giving some time to see if another maintainer can take a look.

PR #754, opened 2026-05-03
fix(pydantic): allow union of BaseModel types as stream_type

### LINKED ISSUE #607: Streaming Event type, type hint, should support union type

Ghraven:
Fixes #607

## What

Broadens the `stream_type` parameter type annotation in `streaming_action.pydantic()` and `pydantic_streaming_action()` to accept a union of pydantic `BaseModel` types in addition to a single `Type[BaseModel]` or `Type[dict]`.

## Why

Users who pass a union of models (e.g. `MyModel1 | MyModel2`) hit a static type error at the call site because the parameter was typed as `Union[Type[BaseModel], Type[dict]]`, which only accepts a single concrete model type. The runtime validation logic in `_validate_and_extract_signature_types_streaming()` already handles union types correctly — this was a type annotation gap only.

## Changes

- `burr/core/action.py`: `stream_type` parameter in `streaming_action.pydantic()` widened to accept `object` (covers union types, which are not expressible as `Type[X]`).
- `burr/integrations/pydantic.py`: Same change to `PartialType` alias and the validator function signature.
- Docstring updated to mention union support.

## Testing

No behaviour change — this is a type annotation fix only. Existing tests continue to pass. Users can now write:

```python
union = MyModel1 | MyModel2

@streaming_action.pydantic(
    reads=[...],
    writes=[...],
    state_input_type=...,
    state_output_type=...,
    stream_type=union,
)
```

without a type error.

### DIFF
--- a/burr/core/action.py
+++ b/burr/core/action.py
@@ -1511,14 +1511,15 @@ class streaming_action:
         writes: List[str],
         state_input_type: Type["BaseModel"],
         state_output_type: Type["BaseModel"],
-        stream_type: Union[Type["BaseModel"], Type[dict]],
+        stream_type: Union[Type["BaseModel"], Type[dict], object],
         tags: Optional[List[str]] = None,
     ) -> Callable:
         """Creates a streaming action that uses pydantic models.
 
         :param reads: The fields this consumes from the state.
         :param writes: The fields this writes to the state.
-        :param stream_type: The pydantic model or dictionary type that is used to represent the partial results.
+        :param stream_type: The pydantic model, dictionary type, or a union of pydantic models
+            (e.g. ``ModelA | ModelB``) used to represent the partial results.
             Use a dict if you want this untyped.
         :param state_input_type: The pydantic model type that is used to represent the input state.
         :param state_output_type: The pydantic model type that is used to represent the output state.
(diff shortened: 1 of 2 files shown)


Ghraven:
Hi! Just checking in on this one. If the union type approach doesn't fit how you'd like to handle this, or if you'd prefer a different fix, I'm happy to adjust. Thanks so much for your work on Burr!

elijahbenizzy:
I don't think loosening it to accept everything is correct? See the comment in the original ticket I just added

Ghraven:
You're absolutely right @elijahbenizzy — loosening to `object` is the wrong shape. `Union[A, B]` isn't a subtype of `A` or `B`, so what's actually needed is to accept union types in the annotation (e.g. `Union[Type[BaseModel], Type[dict], UnionType]`) and have the validator detect the type per-instance at runtime rather than treating the whole union as one model class.

Happy to rework this PR along those lines — adding `UnionType` to the accepted annotations and updating `_validate_and_extract_signature_types_streaming()` (and the pydantic integration validator) to dispatch per-instance when a union is passed. Would that direction work for you? If you'd prefer to leave this one closed and have me open a fresh PR with the correct shape, just say the word. Thanks for the careful review!

PR #762, opened 2026-05-08
Fix "MapStates replays stale sub-state on repeated invocations when parent has a cascading state_initializer"

### LINKED ISSUE #761: MapStates replays stale sub-state on repeated invocations when parent has a cascading state_initializer
When a MapStates (or any MapActionsAndStates / TaskBasedParallelAction) is invoked more than once in the same parent application's lifetime, and the parent application was built with a state_initializer (e.g. via initialize_from(...) for a resumed/forked run), every invocation after the first silently replays the first invocation's sub-state instead of executing the sub-actions. The cause is that sub-application IDs are deterministic in (parent_app_id, i, j) only, so they collide across invocations, and the cascaded initializer then loads the persisted state of the prior call.        
                                                                                                                                                                                                                                                                                                           
  This does not surface on fresh runs (no initialize_from on the parent), because the cascaded initializer is None and sub-apps never load existing state — even though their IDs still collide, each call still runs fresh.

# Reproduction

```python
import hashlib
import random
import tempfile

from burr.core import ApplicationBuilder, State, action
from burr.core.action import create_action
from burr.core.parallelism import MapStates
from burr.tracking.client import LocalTrackingClient


@action(reads=[], writes=["x"])
def pick(state: State) -> State:
    return state.update(x=random.random())


@action(reads=[], writes=[])
def back(state: State) -> State:
    return state


class Fan(MapStates):
    def action(self, state, inputs):
        return create_action(pick, "pick")

    def states(self, state, context, inputs):
        for _ in range(3):
            yield state

    def reduce(self, state, results):
        return state.update(xs=[s["x"] for s in results])

    @property
    def reads(self):
        return []

    @property
    def writes(self):
        return ["xs"]


class FixedFan(Fan):
    """Workaround: mix context.sequence_id into the sub-app id so the per-
    invocation sub-apps get distinct ids and the cascaded initializer can't
    collide them onto a prior call's persisted state."""

    def tasks(self, state, context, inputs):
        for task in super().tasks(state, context, inputs):
            task.application_id = hashlib.sha256(
                f"{task.application_id}:{context.sequence_id}".encode()
            ).hexdigest()
            yield task


def run(fan: MapStates) -> list[list[float]]:
    tracker = LocalTrackingClient(project="test", storage_dir=tempfile.mkdtemp())
    app = (
        ApplicationBuilder()
        .with_actions(fan=fan, back=back)
        .with_transitions(("fan", "back"), ("back", "fan"))
        .with_tracker(tracker)
        .initialize_from(
            tracker,
            resume_at_next_action=True,
            default_state={},
            default_entrypoint="fan",
        )
        .build()
    )
    results = []
    for _ in range(3):
        app.run(halt_after=["fan"])
        results.append(list(app.state["xs"]))
    return results


def main() -> int:
    rc = 0
    for label, fan in [("buggy Fan", Fan()), ("FixedFan", FixedFan())]:
        results = run(fan)
        print(f"\n[{label}]")
        for i, xs in enumerate(results):
            print(f"  invocation {i}: {xs}")
        replayed = all(xs == results[0] for xs in results[1:])
        print(f"  -> {'STALE REPLAY' if replayed else 'fresh outputs'}")
        if (label == "buggy Fan") != replayed:
            rc = 1
    return rc


if __name__ == "__main__":
    raise SystemExit(main())
```

output:

```
❯ uv run repro_mapstates_bug.py
[buggy Fan]
  invocation 0: [0.3601082508430954, 0.47278782334648684, 0.052613734197709316]
  invocation 1: [0.3601082508430954, 0.47278782334648684, 0.052613734197709316]
  invocation 2: [0.3601082508430954, 0.47278782334648684, 0.052613734197709316]
  -> STALE REPLAY

[FixedFan]
  invocation 0: [0.07898183874474951, 0.9919101156042253, 0.5423793059392615]
  invocation 1: [0.5005441699503071, 0.3534622090272228, 0.17871141506633137]
  invocation 2: [0.4519801578765156, 0.07313228344734668, 0.9619878312667838]
  -> fresh outputs
```


Burr 0.40.2

hippalectryon-0:
Fixes #761 

## Changes

Added sequence ID in the tasks keys

## How I tested this

Tested against the MWE of the issue

## Checklist

### DIFF
--- a/burr/core/parallelism.py
+++ b/burr/core/parallelism.py
@@ -518,7 +518,7 @@ class MapActionsAndStates(TaskBasedParallelAction):
         def _tasks() -> Generator[SubGraphTask, None, None]:
             for i, action in enumerate(self.actions(state, context, inputs)):
                 for j, substate in enumerate(self.states(state, context, inputs)):
-                    key = f"{i}-{j}"  # this is a stable hash for now but will not handle caching
+                    key = f"{context.sequence_id}-{i}-{j}"
                     yield _create_task(key, action, substate)
 
         async def _atasks() -> AsyncGenerator[SubGraphTask, None]:
@@ -528,7 +528,7 @@ class MapActionsAndStates(TaskBasedParallelAction):
             states = await async_utils.arealize(state_generator)
             for i, action in enumerate(actions):
                 for j, substate in enumerate(states):
-                    key = f"{i}-{j}"
+                    key = f"{context.sequence_id}-{i}-{j}"
                     yield _create_task(key, action, substate)
 
         return _atasks() if self.is_async() else _tasks()
--- a/tests/core/test_parallelism.py
+++ b/tests/core/test_parallelism.py
@@ -45,7 +45,12 @@ from burr.core.parallelism import (
     _cascade_adapter,
     map_reduce_action,
 )
-from burr.core.persistence import BaseStateLoader, BaseStateSaver, PersistedStateData
+from burr.core.persistence import (
+    BaseStateLoader,
+    BaseStateSaver,
+    InMemoryPersister,
+    PersistedStateData,
+)
 from burr.tracking.base import SyncTrackingClient
 from burr.visibility import ActionSpan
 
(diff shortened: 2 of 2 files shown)


skrawcz (changes requested):
Good catch and clean fix! The root cause analysis in #761 is thorough and the fix is correct — including `sequence_id` in the sub-app key ensures repeated invocations with `initialize_from` get distinct sub-application IDs.

Backwards compatibility is fine here: if old persisted sub-app state exists, `initialize_from` simply falls back to `default_state` and the sub-apps run fresh, which is correct.

One ask before merge: please add a regression test in `tests/core/test_parallelism.py`. The MWE in #761 is a great template — something like:
- Create a `MapStates` action with cascade persistence (`initialize_from` on the parent)
- Run the parent app through the MapStates action twice (two steps)
- Assert the second invocation produces fresh results, not a replay of the first

Without an automated test, a future refactor could easily reintroduce this.

hippalectryon-0:
I checked that the test passes with this PR and fails with the main branch.

PR #771, opened 2026-05-10
website: add Downloads page per ASF release policy

elijahbenizzy:
## Summary

Adds a `/downloads` page to the Burr website covering the
0.42.0-incubating and 0.41.0-incubating releases. Addresses Justin
Mclean's feedback in the 0.42.0-incubating IPMC vote that
burr.apache.org needs a download page per
https://www.apache.org/legal/release-policy.html#publication.

## What's on the page

- **Quick install** — pip install for the convenience wheel
- **Latest release** (0.42.0-incubating, May 9 2026) — source tarball, sdist, and wheel via `closer.lua` mirror selection, with direct HTTPS links to `.asc` signatures and `.sha512` checksums on `downloads.apache.org`
- **Previous releases** (0.41.0-incubating) — same shape; pointer to the ASF archive
- **Verifying releases** — KEYS download, `gpg --verify` and `sha512sum -c` examples, link to the ASF verification doc
- **License & disclaimer** — Apache-2.0 + standard incubator disclaimer

Per ASF policy, only the tarball/wheel artifacts use mirror redirects; the KEYS file, signatures, and checksums are linked HTTPS-direct from `downloads.apache.org`.

## Files

- `website/src/app/downloads/page.tsx` (new)
- `website/src/lib/constants.ts` — adds the "Download" nav link; switches the existing anchor links from `#features`/etc to `/#features` so they still resolve when navigating from `/downloads`
- `website/.gitignore` — adds a negation rule re-including `src/app/downloads/`; the repo-root `.gitignore` has a generic `downloads/` rule (Python packaging convention) that would otherwise hide the new app-router directory

No new dependencies. No styling deviations from the rest of the site (Tailwind tokens + the existing `--card-bg`/`--card-border`/`--muted` color variables).

## Test plan

- [x] `npm run build` succeeds; the new `/downloads` route is statically prerendered (6.19 kB)
- [ ] Visual check on staging: layout, dark/light theme, mobile nav includes the new link
- [ ] Click through every download/signature/checksum link and confirm 200s once 0.42.0-incubating is published to `dist.apache.org`
- [ ] Verify mirror-redirect URLs (`closer.lua`) actually return a mirror after the release lands

PR #774, opened 2026-05-11
ci: gate merging on a single Release Validation summary job

elijahbenizzy:
## Summary

Replaces the two matrix-suffixed required checks (`Release Validation / build-artifacts`, `Release Validation / install-and-smoke (3.12)`) with a single new required check: `Release Validation / summary`.

The new summary job sits at the end of `release-validation.yml`, depends on all the other jobs, and runs with `if: always()`. It translates upstream job conclusions into one binary verdict:

| Upstream `build-artifacts` | Upstream `install-and-smoke` | Summary verdict |
|---|---|---|
| `success` | `success` | ✓ pass |
| `skipped` (path-filtered docs PR) | `skipped` (path-filtered docs PR) | ✓ pass |
| `failure` | * | ✗ fail |
| * | `failure` | ✗ fail |

Because the summary job is the only required check, branch protection cares only about that single stable name.

## Why this is needed

PRs touching only `docs/` or `website/` (like #771) were stuck in `BLOCKED` state. The path filter intentionally skips the heavy validation jobs, but GitHub treats `SKIPPED` as non-passing for required-status-checks. Commit `a655edf2` tried to fix it by adding the `check-paths` gate; that didn't work because `SKIPPED` is still `SKIPPED`.

The `if: always()` summary job sidesteps the ambiguity — it always produces a definite `SUCCESS` or `FAILURE`, never `SKIPPED`. So branch protection has a stable signal regardless of upstream skip semantics.

## What this preserves

Real code PRs still get gated. If `build-artifacts` or `install-and-smoke` fails, the summary fails, and the merge button stays disabled.

## Test plan

- [ ] CI on this PR runs the new summary job and reports green
- [ ] After merge, .asf.yaml updates propagate (a few minutes via ASF infra)
- [ ] #771 (docs-only) flips from `BLOCKED` to `MERGEABLE`

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -62,10 +62,12 @@ github:
       required_status_checks:
         # strict means "Require branches to be up to date before merging".
         strict: false
-        # contexts are the names of checks that must pass
-        contexts:
-          - "Release Validation / build-artifacts"
-          - "Release Validation / install-and-smoke (3.12)"
+        # No required contexts — the Release Validation workflow still runs on
+        # every PR (and on tags) so failures are visible in the checks UI, but
+        # we don't gate merging on it. Path-filtered docs/website PRs were
+        # getting permanently blocked because GitHub treats SKIPPED jobs as
+        # non-passing for required-status-checks.
+        contexts: []
       required_pull_request_reviews:
         dismiss_stale_reviews: false
         require_code_owner_reviews: false


elijahbenizzy:
Going with the soft-pass pattern in the workflow instead — keeps release-validation as a required check and just no-ops cleanly on path-filtered PRs. Replacement coming in a moment.

PR #779, opened 2026-05-16
Fix : Allow streaming actions to yield just state for all but the last one

### LINKED ISSUE #277: Allow streaming actions to yield just state for all but the last one

This should work:

```python
@streaming_action(reads=["prompt"], writes=['response'])
def streaming_response(state: State) -> Generator[dict, None, tuple[dict, State]]:
    response = client.chat.completions.create(
        model='gpt-3.5-turbo',
        messages=[{
            'role': 'user',
            'content': state["prompt"]
            }],
        temperature=0,
    )
    buffer = []
    for chunk in response:
        delta = chunk.choices[0].delta.content
        buffer.append(delta)
        # yield partial results
        yield {'response': delta}
    full_response = ''.join(buffer)
    # return the final result
    return {'response': full_response}, state.update(response=full_response)
```

Instead the yield needs to be `{"response": delta}, None`

If it's not the last yield, we should just yield the result, no need to specify `None` for the state update.

Tasks:
1. Make the change here and in the function below: https://github.com/DAGWorks-Inc/burr/blob/f883885e98a0bdedc1502655ea5fb2ba81f20f49/burr/core/application.py#L253
2. Add tests for this, copy from those that use [this class](https://github.com/DAGWorks-Inc/burr/blob/f883885e98a0bdedc1502655ea5fb2ba81f20f49/tests/core/test_application.py#L555), and corresponding async.

Not doing it

hemanthsavasere:
Fixes : #277 

## Summary
- Allow intermediate yields in `_run_single_step_streaming_action` and `_arun_single_step_streaming_action` to be plain `dict` instead of requiring `(dict, None)` tuples
- Backward compatible — existing `(dict, None)` yields still work
- Added callback tests verifying PostStreamItemHook fires correctly for plain dict yields
- Updated `streaming_action` docstring example to show plain dict yield

### DIFF
--- a/burr/core/action.py
+++ b/burr/core/action.py
@@ -1545,8 +1545,8 @@ class streaming_action:
         See the following example for how to use this decorator -- this reads ``prompt`` from the state and writes
         ``response`` back out, yielding all intermediate chunks.
 
-        Note that this *must* return a value. If it does not, we will not know how to update the state, and
-        we will error out.
+        Note that this *must* return a final value with a state update. If it does not, we will not know how to update the state, and
+        we will error out. Intermediate yields can be plain dicts (without a state update).
 
         .. code-block:: python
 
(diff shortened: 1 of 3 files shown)


skrawcz (approved):
Clean ergonomic improvement — well-tested and backwards compatible. The existing `state_update is None` guard after generator exhaustion ensures users still get a clear error if they forget the final `(result, state)` tuple.

Nice first contribution!

PR #789, opened 2026-05-24
fix: use utf-8 for test case json files

Ghraven:
## Summary
- read and write generated test-case JSON files with explicit UTF-8 encoding
- read dynamic pytest test-case JSON fixtures with the same encoding

## Before / after
Before, the CLI test-case generator and `load_test_cases()` relied on the platform default text encoding for JSON fixture files.
After, both paths use UTF-8 explicitly, making generated fixtures portable across non-UTF-8 locales.

## Verification
- `python -m py_compile burr/cli/__main__.py burr/testing/__init__.py`
- `git diff --check`

### DIFF
--- a/burr/cli/__main__.py
+++ b/burr/cli/__main__.py
@@ -477,14 +477,14 @@ def create_test_case(
     if target_file_name:
         # if it already exists, load it up and append to it
         if os.path.exists(target_file_name):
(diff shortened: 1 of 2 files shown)


andreahlert (approved):
LGTM following #790

PR #790, opened 2026-05-26
fix: use utf-8 for local tracking text files

Ghraven:
## Summary

Thanks for maintaining Burr. This PR makes the local tracking backend text-file reads and writes use explicit UTF-8 encoding.

## What changed

- Adds `encoding="utf-8"` to annotation JSONL reads and writes in `LocalBackend`
- Adds `encoding="utf-8"` when reading `children.jsonl`
- Leaves binary log/metadata reads unchanged

## Before / after

Before, these JSONL text files used the platform default encoding, which can vary on Windows and other non-UTF-8 locale environments.

After, the local backend reads and writes these text files consistently as UTF-8.

## Verification

- `python -m py_compile burr/tracking/server/backend.py`
- `git diff --check`

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -69,7 +69,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install -e ".[tests,tracking-client,graphviz]"
+          python -m pip install -e ".[tests,tracking-client,tracking-server,graphviz]"
 
       - name: Run tests
         run: |
--- a/burr/tracking/server/backend.py
+++ b/burr/tracking/server/backend.py
@@ -309,7 +309,7 @@ class LocalBackend(BackendBase, AnnotationsBackendMixin):
         annotations_path = self._get_annotation_path(project_id)
         annotations = []
         if os.path.exists(annotations_path):
-            async with aiofiles.open(annotations_path) as f:
+            async with aiofiles.open(annotations_path, encoding="utf-8") as f:
                 for line in await f.readlines():
                     annotations.append(AnnotationOut.parse_raw(line))
         return annotations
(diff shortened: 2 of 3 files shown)


andreahlert (changes requested):
Thanks for the fix. One spot in the same file looks missed: `backend.py:524` reads `graph.json` text-mode without `encoding="utf-8"`. Same class of bug, and client writes it as UTF-8 (`client.py:420`).

```python
async with aiofiles.open(graph_file, encoding="utf-8") as f:
```

Could you include that line in this PR as well?

Ghraven:
Thank you for catching that missed spot. I added the same explicit `encoding="utf-8"` to the `graph.json` read in `LocalBackend` and pushed the update. I verified it with `python -m py_compile burr/tracking/server/backend.py` and `git diff --check`.

skrawcz:
This looks like the right fix, but since this is correcting platform-dependent file encoding behavior, I think it would be worth adding a regression test that writes and reads non-ASCII content through LocalBackend (annotations / graph / children) to prove UTF-8 handling end-to-end and prevent missed call sites. Could you add that please?

Ghraven:
Thank you for the suggestion. I added a focused regression test that writes non-ASCII content through LocalBackend annotations, graph.json, and children.jsonl, then reads it back through the backend APIs.\n\nVerified locally with:\n\n``\npython -m pytest tests\\tracking\\test_local_tracking_client.py::test_local_backend_reads_utf8_annotations_graph_and_children -q\npython -m py_compile burr\\tracking\\server\\backend.py tests\\tracking\\test_local_tracking_client.py\ngit diff --check\n``\n\nThe focused pytest passes; the remaining warnings are existing Pydantic deprecation warnings in the backend path.

andreahlert (approved):
LGTM. Thanks for adding the test and catching the graph.json spot.

I pushed one small follow-up commit (b142647d) to harden the regression test. The round-trip assertion alone passes on a UTF-8 default host even if a text open drops encoding="utf-8", so it would not actually catch a future missed call site on CI. The follow-up wraps the backend's aiofiles.open and asserts every text-mode open is explicitly UTF-8 (binary rb opens stay exempt), which fails deterministically regardless of the host locale. Verified it goes red when the encoding arg is removed from any read.

Merging once CI is green.

PR #794, opened 2026-06-02
Add optional module allowlist to pydantic serde deserializer

andreahlert:
This change introduces a configurable allowlist for pydantic deserialization to give users more control over which modules can be dynamically loaded during state restoration.

Changes:
- `set_allowlist([...])` for global configuration
- Per-call `allowlist=` kwarg in `State.deserialize()`
- Backward-compatible: without an allowlist, behavior is unchanged with a runtime warning
- Rejects unauthorized modules with a clear `ValueError`

Includes tests for allowlist acceptance, rejection, and global configuration.


jernejfrank (approved):
Great!

PR #796, opened 2026-06-02
feat(core): exception transitions via on_error capture and wildcard routing (#30)

### LINKED ISSUE #30: Add exception transitions
Currently, exceptions will break the control flow of an action, stopping the program early. Thus, if an exception is expected, the program will stop early. We will be adding the ability to conditionally transition based on exceptions, which will allow you to transition to an error-handling (or retry) action that does not need the full outputs of the prior action.

Here is what it would look liek in the current API:

```python
@action(reads=["attempts"], writes=["output", "attempts"])
def some_flaky_action(state: State, max_retries: int=3) -> Tuple[dict, State]:
    result = {"output": None, "attempts": state["attempts"] + 1}
    try:
        result["output"] = call_some_api(...)
    excecpt APIException as e:
        if state["attempts"] >= max_retries:
           raise e
    return result, state.update(**result)
```

One could imagine adding it as a condition (a few possibilities)
```python
@action(reads=[], writes=["output"])
def some_flaky_action(state: State) -> Tuple[dict, State]:
    result = {"output": call_some_api(...)}
    return result, state.update(**result)

builder.with_actions(
   some_flaky_action=some_flaky_action
).with_transitions(
   (
      "some_flaky_action",
      "some_flaky_action",
      error(APIException) # infinite retries
      error(APIException, max=3) # 3 visits to this edge then it gets reset if this is not chosen
      # That's stored in state
)
```

andreahlert:
## Summary

Implements [#30](https://github.com/apache/burr/issues/30) "Add exception transitions". Currently an exception in an action breaks control flow and stops the program. This adds the ability to **suppress an exception, capture it into state, and route to an error-handling action** via a wildcard transition.

Scope chosen: **capture + wildcard** (combines ideas #3 and #4 from the issue thread).

```python
from burr.core import action, capture_as, expr, ApplicationBuilder, State

@action(reads=[], writes=["output"], on_error=capture_as("error"))
def flaky(state: State) -> tuple[dict, State]:
    result = {"output": call_some_api(...)}
    return result, state.update(**result)

@action(reads=["error"], writes=[])
def handler(state: State) -> tuple[dict, State]:
    ...  # inspect / reset state["error"], then route onward

app = (
    ApplicationBuilder()
    .with_actions(flaky=flaky, handler=handler)
    .with_state(error=None)  # seed: expr() raises on a missing key
    .with_transitions(("*", "handler", expr("error is not None")))
    .with_entrypoint("flaky")
    .build()
)
```

## What's included

- **`capture_as(field, include_traceback=True)`** — an `on_error` handler that writes a JSON-serializable `{type, message, traceback}` record into a state field. Exported from `burr.core`.
- **`@action(..., on_error=handler)`** — per-action handler, any callable `(State, Exception) -> State`.
- **`ApplicationBuilder.with_error_handling(handler)`** — builder-level global handler for actions without their own `on_error`. Per-action takes precedence.
- **Wildcard `("*", target, condition)` transitions** — route from any action.

The captured field bypasses reducer write-validation (need not be in `writes`). A failing handler never masks the original exception (re-raised with the handler error as cause). Captured records are JSON-serializable, so persistence/tracking are unaffected.

## Behavior changes to call out for review

These go beyond the headline feature, kept as separate commits:

1. **4-tier transition resolution** (`fix(core)` commit). The fix for the precedence bug that kept #30 open: an action's own `default` transition shadowed a guarded wildcard route, so the route never fired. `get_next_node` now resolves: source non-default, wildcard non-default, source default, wildcard default. **Semantic change**: in default-FIRST graphs a previously-dead non-default transition becomes live. Default-LAST graphs (the common case) and graphs without wildcards are unchanged, covered by a regression test.
2. **`_astep` restructure**. A SYNC action driven through the async path now reports the real `result`/`state` to the async `post_run_step` hook instead of stale values (it previously delegated to `_step` and discarded them). Locked by a regression test.

## Out of scope (follow-up)

Streaming error handling (`@streaming_action` `on_error` during incremental `stream_result`/`astream_result` consumption) is intentionally not included, to avoid shipping a silently inert API. Tracked as a follow-up on #30.

## Testing

- `pytest tests/core/` green (363 tests, +17 new).
- Full non-integration suite green (416 passed, 1 skipped).
- Verified wildcard graphs survive `visualize()` and `ApplicationModel` serialization (the tracking UI path).
- `flake8` clean, `black`/`isort` applied.

## Process

The design never converged on the list (the issue thread has 4 competing shapes). This PR is a concrete proposal. I will start a `[DISCUSS]` thread on dev@burr.apache.org referencing this branch before merge, per podling process. Feedback on the precedence semantics in particular is welcome.

Closes #30.

### DIFF
--- a/burr/core/action.py
+++ b/burr/core/action.py
@@ -22,6 +22,7 @@ import copy
 import inspect
 import sys
 import textwrap
+import traceback
 import types
 import typing
 from collections.abc import AsyncIterator
@@ -378,6 +379,82 @@ class Action(Function, Reducer, abc.ABC):
         :return: List of string tags
         """
         return []
+
+    @property
+    def on_error(self) -> Optional[Callable[[State, Exception], State]]:
+        """Returns the error handler associated with this action, if any.
+
+        An error handler is any callable ``(State, Exception) -> State``. If it is set
+        and the action raises, the application will call it to produce a new state
+        (suppressing the exception) instead of re-raising. See
+        :py:func:`capture_as <burr.core.action.capture_as>` for a built-in handler.
+
+        :return: The error handler callable, or None if not set
+        """
+        return getattr(self, "_on_error", None)
+
+
+class capture_as:
+    """Error handler that suppresses an exception and records a JSON-serializable
+    summary of it into the given state field.
+
+    Use this with the ``on_error`` parameter of the :py:func:`@action <burr.core.action.action>`
+    decorator, or with :py:meth:`ApplicationBuilder.with_error_handling
+    <burr.core.application.ApplicationBuilder.with_error_handling>`:
+
+    .. code-block:: python
+
+        @action(reads=[], writes=["output"], on_error=capture_as("error"))
+        def flaky(state: State) -> tuple[dict, State]:
+            result = {"output": call_some_api(...)}
+            return result, state.update(**result)
+
+        @action(reads=["error"], writes=[])
+        def handler(state: State) -> tuple[dict, State]:
+            ...  # inspect/reset state["error"], then route onward
+
+        app = (
+            ApplicationBuilder()
+            .with_actions(flaky=flaky, handler=handler)
+            # the capture field MUST be initialized -- expr() raises on a missing key,
+            # so seed ``error`` to None so the wildcard condition is safe to evaluate
+            # on every step before any exception has occurred.
+            .with_state(error=None)
+            .with_transitions(("*", "handler", expr("error is not None")))
+            .with_entrypoint("flaky")
+            .build()
+        )
+
+    Note the captured field need NOT appear in the action's ``writes`` -- the handler
+    writes it directly, bypassing reducer write-validation. When the action raises,
+    ``state["error"]`` is set to a JSON-serializable dict like::
+
+        {"type": "ValueError", "message": "...", "traceback": "..."}
+
+    which the wildcard transition above then routes on.
+    """
+
+    def __init__(self, field: str, include_traceback: bool = True):
+        """:param field: State field to write the error record to.
+        :param include_traceback: Whether to include the formatted traceback string.
+        """
+        self.field = field
+        self.include_traceback = include_traceback
+
+    def __call__(self, state: State, exception: Exception) -> State:
+        record = {"type": type(exception).__name__, "message": str(exception)}
+        if self.include_traceback:
+            # Formatting must never fail -- a handler that raises would mask the
+            # original exception in the calling context. Fall back to a safe
+            # placeholder if traceback formatting blows up (e.g. corrupted
+            # traceback objects or pathological custom exceptions).
+            try:
+                record["traceback"] = "".join(
+                    traceback.format_exception(type(exception), exception, exception.__traceback__)
+                )
+            except Exception as e:  # noqa: BLE001
+                record["traceback"] = f"<Traceback formatting failed: {e}>"
+        return state.update(**{self.field: record})
 
 
 class Condition(Function):
@@ -806,6 +883,7 @@ class FunctionBasedAction(SingleStepAction):
         originating_fn: Optional[Callable] = None,
         schema: ActionSchema = DEFAULT_SCHEMA,
         tags: Optional[List[str]] = None,
+        on_error: Optional[Callable[[State, Exception], State]] = None,
     ):
         """Instantiates a function-based action with the given function, reads, and writes.
         The function must take in a state and return a tuple of (result, new_state).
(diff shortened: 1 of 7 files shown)


burr/core/action.py
```diff
@@ -379,6 +380,82 @@ def tags(self) -> list[str]:
...
+        app = (
+            ApplicationBuilder()
+            .with_actions(flaky=flaky, handler=handler)
+            # the capture field MUST be initialized -- expr() raises on a missing key,
+            # so seed ``error`` to None so the wildcard condition is safe to evaluate
+            # on every step before any exception has occurred.
+            .with_state(error=None)
+            .with_transitions(("*", "handler", expr("error is not None")))
+            .with_entrypoint("flaky")
+            .build()
```
jernejfrank, line 425: Trying to understand this better. This means we need to populate something in the state, how does this play with existing states that are already persisted and the builder loads them in? Is it enough to seed the last state and turn on the new error handling API or do we need to retrospectively populate all of them?

jernejfrank: Poking around a bit to see if I understand this.

If I need to handle different errors and I route them to different handlers, then with this API I will need to specify each one individually?

```python
app = (
            ApplicationBuilder()
            .with_actions(flaky=flaky, handler=handler)
            # the capture field MUST be initialized -- expr() raises on a missing key,
            # so seed ``error`` to None so the wildcard condition is safe to evaluate
            # on every step before any exception has occurred.
            .with_state(error1=None,error2=None,error3=None,...)
            .with_transitions(
                ("error1", "handler", expr("error1 is not None"),
                ("error2", "handler", expr("error2 is not None"),
                ("error3", "handler", expr("error3 is not None"),
            )
            .with_entrypoint("flaky")
            .build()
```

burr/core/application.py
```diff
@@ -849,6 +849,7 @@ def __init__(
         parallel_executor_factory: Optional[Executor] = None,
         state_persister: Union[BaseStateSaver, LifecycleAdapter, None] = None,
         state_initializer: Union[BaseStateLoader, LifecycleAdapter, None] = None,
+        global_error_handler: Optional[Callable[[State, Exception], State]] = None,
```
jernejfrank, line 852: Digging into this a bit:

IIUC, each action can specify their own error type, state dict key for the error,m etc.. But that can get tedious if you have multiple actions that need the same error handling. So this gives a global handling ability, but it does it only for a single type of error? Like, I would not be able to specify globally a sequence of handlers that handle different errors, unless I wrap that inside the `global_error_handler` callable, make that thing cathc the umbrella exception and than inside continue routing to subhandlers?

burr/core/graph.py
```diff
@@ -59,7 +59,9 @@ def _validate_transitions(
 ):
     exhausted = {}  # items for which we have seen a default transition
     for from_, to, condition in transitions:
-        if from_ not in actions:
+        # "*" is a wildcard source -- it routes from ANY action, so it is not
+        # required to be a declared action. The target, however, must be real.
+        if from_ != "*" and from_ not in actions:
```
jernejfrank, line 64: Nit: this may sound really pedantic but I would prefer `"*"` to be either enum or global variable

burr/core/graph.py
```diff
@@ -150,13 +154,36 @@ def _create_action_tag_map(actions: List[Action]) -> dict[str, List[Action]]:
...
+        tiers = (
+            [(t, c) for t, c in src if not _is_default(c)],
+            [(t, c) for t, c in wild if not _is_default(c)],
+            [(t, c) for t, c in src if _is_default(c)],
+            [(t, c) for t, c in wild if _is_default(c)],
+        )
+        for tier in tiers:
+            for next_action, condition in tier:
+                if condition.run(state)[Condition.KEY]:
+                    return self._action_map[next_action]
```
jernejfrank, line 186: Can this blow up in our faces for non-error nodes if somebody gets funky with their naming?

jernejfrank (commented):
Hey, this is a really cool feature and much needed. Couple of thoughts from my side already below.

1. The one think I am not completely getting is the intended usage. We have two APIs, the per action one which takes precedence, and a global fallback. They seem to differ in the scope and usage tho (unless I am getting something wrong):

  a. `on_error=capture_as("error")` Does strict just dump the error as a string into the state. The handler for that than needs to be a node in the graph to do something more? But technically each action could have a different error string and an associated error handler node.
  b. `global_error_handler`, you can set on the builder and pass in directly a python function to handle it for you. Here you decide what to do with it since you have control of the callable -- can potentially even forget to save it state -- and do not need a node in the graph to do the extra handling. BUT, you can only use a single handler and cannot differentiate between different exceptions.

2. Maybe it is just me, but I would prefer to roll this out in one place first (either globally via the builder or as the action API) to test observe how it behaves in the wild before expanding too much API surface. 

3. String handling can get really annoying and tedious. It also prohibits any inheritance cascading (which I guess is the reason for the wildcard). Would it be better to do error handling based directly on Python `Exception` types? IMO, this would unlock also that handlers can be defined hierarchical and mimick Python std exception handling which should be familiar to users?

4. With the above innate error handling I feel like this can be super simplified: You just need to define Burr nodes that handle exceptions -- basically this would promote the try-except python code on the level of Burr nodes. Basically every node is a try, an dedicated Burr nodes or builder handlers act as the except clauses you have in normal python.

PR #818, opened 2026-06-22
Add optional allowlist to pickle serde deserializer

elijahbenizzy:
## Summary

Adds an optional allowlist to the pickle deserializer used by Burr's serde system (`burr/integrations/serde/pickle.py`), mirroring the shape of #794's pydantic allowlist work.

Without an allowlist: backward-compatible — `deserialize_pickle` continues to call `pickle.loads(...)` and emits a one-time `SecurityWarning` per registration site so users see the noise.

With an allowlist: a `_RestrictedUnpickler(pickle.Unpickler)` subclass overrides `find_class(module, name)` to validate against the allowlist and raises `pickle.UnpicklingError` for anything not on it.

## API

Three ways to set the allowlist, in resolution order (highest priority first):

1. Per-call kwarg: `deserialize_pickle(value, allowlist=[...])`
2. At registration time: `register_type_to_pickle(cls, allowlist=[...])`
3. Process-wide: `set_pickle_serde_allowlist([...])`
4. Otherwise: legacy unsafe path + `SecurityWarning`

Allowlist entries are `(module, qualname)` tuples — e.g. `("myapp.models", "User")` — rather than prefix strings. Pickle attacks routinely reach for *specific symbols within otherwise-trusted modules* (e.g. `builtins.eval`, `os.system`, `subprocess.Popen`), so per-class granularity matters more than for pydantic where the registered name is already a class.

## Trust model (now in the docstring)

Pickle deserialization is unsafe when the bytes come from a source the application doesn't fully control — including persistence backends that have a separate access model (SQLite files, Redis, S3, filesystems). An attacker with write access to the backend can plant a malicious `__reduce__` payload that triggers code execution when state is loaded. The allowlist closes that primitive.

## Tests

8 tests in `tests/integrations/serde/test_pickle.py` (was 1):

- `_is_allowed` truth table for allowlist matching
- Malicious `__reduce__` payload (calling a sentinel function via `__reduce__`) is blocked when an allowlist is set; sentinel never runs
- Legitimate object roundtrips when its `(module, qualname)` is on the allowlist
- Module-level `set_pickle_serde_allowlist` applies to fresh registrations
- Instance-level allowlist (at `register_type_to_pickle`) overrides module default
- Per-call allowlist kwarg overrides both
- Legacy path (no allowlist) still works and emits the `SecurityWarning` exactly once per registration site

Wider `tests/integrations/serde/` + `tests/core/test_serde.py` runs 24/24 green.

### DIFF
--- a/burr/integrations/serde/pickle.py
+++ b/burr/integrations/serde/pickle.py
@@ -18,12 +18,94 @@
 # Pickle serde registration
 # This is not automatically registered because we want to register
 # it based on class type.
+import io
 import pickle
+import warnings
+from typing import Iterable, List, Optional, Tuple
 
 from burr.core import serde
 
+# Type alias: an allowlist entry is ``(module, qualname)``.
+# ``qualname`` may be ``"*"`` to permit any class in the given module.
+AllowlistEntry = Tuple[str, str]
 
-def register_type_to_pickle(cls):
+
+class SecurityWarning(Warning):
+    """Warning issued when pickle deserialization proceeds without an allowlist."""
+
+
+# Global allowlist for pickle deserialization. When set, only
+# ``(module, qualname)`` pairs in this list are permitted to be
+# reconstructed during ``pickle.loads``.
+_global_allowlist: Optional[List[AllowlistEntry]] = None
+
+
+def set_pickle_serde_allowlist(allowlist: Optional[Iterable[AllowlistEntry]]) -> None:
+    """Set a process-wide allowlist of ``(module, qualname)`` pairs that may be
+    reconstructed by ``deserialize_pickle``.
+
+    Each entry is a tuple ``(module_name, class_name)``. Use ``"*"`` as the class
+    name to permit any class in that module:
+
+    .. code-block:: python
+
+        from burr.integrations.serde import pickle as burr_pickle
+
+        burr_pickle.set_pickle_serde_allowlist([
+            ("myapp.models", "User"),
+            ("myapp.models", "Address"),
+            ("myapp.types", "*"),  # any class in myapp.types
+        ])
+
+    When the allowlist is set, any attempt to deserialize a class outside the
+    allowlist raises ``pickle.UnpicklingError``. Pass ``None`` to clear it.
+
+    :param allowlist: Iterable of ``(module, qualname)`` tuples, or ``None``.
+    """
+    global _global_allowlist
+    _global_allowlist = list(allowlist) if allowlist is not None else None
+
+
+def _is_allowed(module: str, name: str, allowlist: Iterable[AllowlistEntry]) -> bool:
+    for allowed_module, allowed_name in allowlist:
+        if module == allowed_module and (allowed_name == "*" or allowed_name == name):
+            return True
+    return False
+
+
+class _RestrictedUnpickler(pickle.Unpickler):
+    """Pickle unpickler that only resolves classes present in an allowlist.
+
+    See https://docs.python.org/3/library/pickle.html#restricting-globals
+    for the standard library guidance this is modeled on.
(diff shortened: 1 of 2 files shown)


burr/integrations/serde/pickle.py
```diff
@@ -18,12 +18,94 @@
...
+    return _RestrictedUnpickler(io.BytesIO(data), allowlist, **kwargs).load()
+
+
+# Tracks call sites that have already received a "no allowlist" warning, so the
+# warning is emitted at most once per registration site rather than once per
+# deserialization (which could be noisy on hot paths).
+_warned_call_sites: set = set()
+
+
+def register_type_to_pickle(cls, allowlist: Optional[Iterable[AllowlistEntry]] = None):
```
andreahlert, line 108: this is fail-open. key's static `"pickle"`, so registering another class w/o a list later wipes this one's allowlist. tested it. lvl-2 should just go — per-call + global is enough (#794 already does it that way).

burr/integrations/serde/pickle.py
```diff
@@ -18,12 +18,94 @@
...
+    """Warning issued when pickle deserialization proceeds without an allowlist."""
+
+
+# Global allowlist for pickle deserialization. When set, only
+# ``(module, qualname)`` pairs in this list are permitted to be
+# reconstructed during ``pickle.loads``.
+_global_allowlist: Optional[List[AllowlistEntry]] = None
+
+
+def set_pickle_serde_allowlist(allowlist: Optional[Iterable[AllowlistEntry]]) -> None:
```
andreahlert, line 43: two setters + two SecurityWarnings now (this and pydantic's). worth merging into one `burr.serde` someday?

burr/integrations/serde/pickle.py
```diff
@@ -40,9 +122,40 @@ def __init__(self, name, email):
         from burr.integrations.serde import pickle
         pickle.register_type_to_pickle(User) # this will register the User class to be serialized/deserialized using pickle.
 
+    Trust model
+    -----------
+    Pickle is, by design, capable of executing arbitrary code during
+    deserialization. If the persistence backend (SQLite file, Redis, S3, the
+    local filesystem, etc.) can be written to by an untrusted party, a tampered
+    payload can trigger remote code execution when burr restores application
+    state.
```
andreahlert, line 131: restricting globals isn't a full fix (cpython docs say so) — softening this + mentioning signed payloads:

```suggestion
    state.

    Note: an allowlist restricts *which* globals can be imported — necessary
    but not sufficient (see CPython "Restricting Globals"). For fully untrusted
    backends, prefer not unpickling, or sign payloads (HMAC) as defense in depth.
```

tests/integrations/serde/test_pickle.py
```diff
@@ -25,22 +31,176 @@ def __init__(self, name, email):
...
+
+def test_is_allowed_logic():
+    assert _is_allowed("m", "C", [("m", "C")]) is True
+    assert _is_allowed("m", "C", [("m", "*")]) is True
+    assert _is_allowed("m", "C", [("other", "*")]) is False
+    assert _is_allowed("m", "C", [("m", "D")]) is False
+    assert _is_allowed("m", "C", []) is False
+
+
+def test_malicious_reduce_payload_blocked_with_allowlist():
```
andreahlert, line 71: only catches a top-level `builtins.int`. one where an allowed obj nests a blocked class would be a stronger test.

andreahlert (changes requested):
Thanks for this — solid direction (`find_class` is the right call, back-compat's clean). Just a few notes: only the first one's a blocker (the reg-time allowlist is fail-open), the rest are nits.

PR #820, opened 2026-06-22
Add Condition.safe_expr for opt-in restricted AST evaluation

### LINKED ISSUE #817: Provide opt-in safe AST evaluator for Condition.expr() to support untrusted expression sources
## Background

`Condition.expr(expr_string)` in `burr/core/action.py` is the compact-DSL way to express a transition guard against state — e.g. `expr("step_count > 10")`. Today it parses the string with `ast.parse()`, compiles it, and `eval()`s it against `state.get_all()` as locals.

This is a deliberately full-power evaluator: it accepts arbitrary Python in the expression. The docstring already notes:

> *"Do not accept expressions generated from user-inputted text."*

That contract is appropriate for the **intended use case** — the framework author writes the expression as part of the graph definition, the same way they could write a lambda. There's no plan to change the default behavior.

## The gap

There's a real second use case where users want to accept expressions from a less-trusted source (e.g. a dashboard config, a SaaS user-facing rule editor, a YAML-driven graph definition). For that case, Burr currently offers no safe surface — they'd have to wrap `expr()` in their own AST-validating layer.

## Proposal

Provide an **opt-in** safe AST evaluator as a separate entry point. Sketch:

- `Condition.safe_expr(expr_string)` — new public API, restricted-AST evaluator.
- Implementation walks the parsed `ast.AST` with a `NodeVisitor` and rejects anything not on an allowlist.
- Allowed node types (rough cut, to refine in design):
  - Constants (`int`, `float`, `str`, `bool`, `None`)
  - Names (resolved against state)
  - Attribute access on Names (but reject any `__dunder__` attribute access)
  - Subscript access (`state["key"]`, `items[0]`)
  - Comparisons (`==`, `!=`, `<`, `>`, `<=`, `>=`, `in`, `not in`, `is`, `is not`)
  - Boolean ops (`and`, `or`, `not`)
  - Arithmetic on numbers (`+`, `-`, `*`, `/`, `//`, `%`)
  - Tuple/list/set/dict literals
  - Calls to a small allowlist of safe builtins: `len`, `abs`, `min`, `max`, `sum`, `all`, `any`, `str`, `int`, `float`, `bool` (return-typed, no side effects)
- Explicitly rejected:
  - Any other `Call` node
  - `Lambda`, `IfExp`, comprehensions (`ListComp`, `DictComp`, `SetComp`, `GeneratorExp`)
  - `Import`/`ImportFrom`
  - `Subscript` slicing with `Slice` if we want to be conservative (TBD)
  - Walrus (`NamedExpr`) — TBD
  - Dunder attribute access on any node

Evaluation strategy: walk the validated tree and interpret directly, rather than calling `eval()` on the validated source. This removes the dependency on CPython's `eval()` semantics entirely.

## Out of scope for this issue
- Changing the default behavior of `Condition.expr()`. It stays as today (full-power, author-trusted).
- Adding a new dependency (e.g. `simpleeval`, `asteval`). The implementation should be in-tree.

## Acceptance criteria
- New `safe_expr` (or equivalently named) public API on `Condition`.
- Existing test patterns for `expr()` ported to `safe_expr` to confirm the common usage works.
- Negative tests proving each rejected node category raises a clear `SyntaxError` / `ValueError` at parse time, not at eval time.
- Docstring on `expr()` updated to clearly contrast the two surfaces.
- Worth a CHANGELOG entry as a feature.

## Rough sizing
Focused module of new code, defined contract, no live design choices once node-allowlist is settled — ~300–600 LOC including tests. Reviewable in ~20 minutes by someone who knows the AST module.

elijahbenizzy:
Closes #817.

## Summary

Adds a new public API `Condition.safe_expr()` (and a top-level `safe_expr` re-export from `burr.core` for symmetry with `expr`) that evaluates a string expression under a strict AST allowlist, intended for callers that want to accept expression strings from less-trusted sources (dashboard rule editors, YAML-driven graph definitions, configuration files).

Existing `Condition.expr()` is **unchanged** — it remains a developer-authored, full-power eval surface. The trust-model contrast is captured in both functions' docstrings, with `expr()` now pointing readers at `safe_expr()` for the untrusted-input case (cross-reference docstring update lands in this PR).

## Safety argument

The validated AST is **interpreted directly** — there is no `eval()` or `compile()` call on the tree. The pipeline is:

1. `ast.parse(expr, mode="eval")` (parse only — does not execute).
2. `_SafeExprValidator(ast.NodeVisitor)` walks the tree. `generic_visit` is a hard-rejection catch-all: any node type not explicitly opted in raises `ValueError` at *parse/validate time*, not at runtime.
3. `_SafeExprInterpreter` walks the validated tree via per-node-type methods and produces a value.
4. Builtins available inside expressions are looked up in a hard-coded 11-name dict (`_SAFE_EXPR_BUILTINS`), not `builtins.__dict__`.

Because no `eval()` is ever invoked, the safety surface does not depend on CPython's `eval` semantics (no `__builtins__` auto-injection concern, no globals/locals juggling).

## What's allowed

Constants; `Name`; `Attribute` access (but **any** `__dunder__` attribute access is rejected — closes the `(0).__class__.__bases__[0].__subclasses__()` escape); `Subscript`; `Compare` (full set including `Is`/`IsNot`/`In`/`NotIn`); `BoolOp` (`and`/`or`); `UnaryOp` (`Not`, `USub`, `UAdd`); arithmetic `BinOp` (`+ - * / // % **`); literal containers (`Tuple`, `List`, `Set`, `Dict`); `Call` only to the allowlist `{len, abs, min, max, sum, all, any, str, int, float, bool}`.

## What's rejected at parse time

`Lambda`, `IfExp`, all comprehensions, `Import`/`ImportFrom`, `Yield`/`YieldFrom`/`Await`, `NamedExpr` (walrus), any `Call` to a non-allowlisted name, any `Attribute` access starting with `__`. Anything not explicitly opted in falls through to `generic_visit` which raises.

## Tests

47 new tests in `tests/core/test_action.py` (≈19 positive — grammar coverage, builtins, determinism, edge cases — and ≈28 negative, including 7 explicit attack-pattern tests like `__import__("os").system(...)`, `(0).__class__.__bases__[0].__subclasses__()`, `open("/etc/passwd").read()`, `lambda: 1`, comprehensions).

Full test file: **154 passed in 1.37s**.

## Open questions for reviewers

1. **Builtin allowlist** is strict per the issue — `round`, `sorted`, `set`/`list`/`tuple`/`dict` constructors, `range` are intentionally *not* included. `range` in particular is hazardous (DoS via `sum(range(10**12))`). Worth confirming this is the desired posture for the initial cut.
2. **Attribute access** is allowed broadly on non-dunder names. A future opt-out (`safe_expr(..., allow_attributes=False)`) for stricter deployments would be a small add.
3. **`Is` / `IsNot`** are allowed per the issue but are a string/int-identity footgun. Worth flagging in the docstring or leaving for a future hardening pass.
4. **Resource bounds** — no timeout, depth, or memory limits. `safe_expr("x ** y ** y")` with attacker-controlled `x, y` can still hang or OOM. Out of scope for this PR; worth a follow-up issue.
5. **`safe_when` (kwarg form)** symmetric to `safe_expr` — not in the issue. Flagging for triage.

### DIFF
--- a/burr/core/__init__.py
+++ b/burr/core/__init__.py
@@ -15,7 +15,17 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from burr.core.action import Action, Condition, Result, action, default, expr, type_eraser, when
+from burr.core.action import (
+    Action,
+    Condition,
+    Result,
+    action,
+    default,
+    expr,
+    safe_expr,
+    type_eraser,
+    when,
+)
 from burr.core.application import (
     Application,
     ApplicationBuilder,
@@ -35,6 +45,7 @@ __all__ = [
     "Condition",
     "default",
     "expr",
+    "safe_expr",
     "type_eraser",
     "Result",
     "State",
--- a/burr/core/action.py
+++ b/burr/core/action.py
@@ -380,6 +380,320 @@ class Action(Function, Reducer, abc.ABC):
         return []
 
 
+# ---------------------------------------------------------------------------
+# Safe expression evaluator used by Condition.safe_expr.
+#
+# These are intentionally module-private. The validator walks the AST and
+# rejects any node that isn't on the allowlist; the interpreter then walks
+# the (validated) tree and produces a value. eval()/compile() are never
+# called on the parsed tree -- that is the entire point.
+# ---------------------------------------------------------------------------
+
+# Builtins we are willing to expose inside safe_expr. Everything here returns
+# a value (no side effects), takes only basic Python data, and cannot be used
+# to reach the import system or the interpreter internals.
+_SAFE_EXPR_BUILTINS: typing.Dict[str, Callable] = {
+    "len": len,
+    "abs": abs,
+    "min": min,
+    "max": max,
+    "sum": sum,
+    "all": all,
+    "any": any,
+    "str": str,
+    "int": int,
+    "float": float,
+    "bool": bool,
+}
+
+# AST BinOp / UnaryOp / BoolOp / Compare operator classes we accept.
+_SAFE_BINOPS = (
+    ast.Add,
+    ast.Sub,
+    ast.Mult,
+    ast.Div,
+    ast.FloorDiv,
+    ast.Mod,
+    ast.Pow,
+)
+_SAFE_UNARYOPS = (ast.Not, ast.USub, ast.UAdd)
+_SAFE_BOOLOPS = (ast.And, ast.Or)
+_SAFE_CMPOPS = (
+    ast.Eq,
+    ast.NotEq,
+    ast.Lt,
+    ast.LtE,
+    ast.Gt,
+    ast.GtE,
+    ast.In,
+    ast.NotIn,
+    ast.Is,
+    ast.IsNot,
+)
+
+
+class _SafeExprValidator(ast.NodeVisitor):
+    """Walks an AST and raises ``ValueError`` on any node not on the allowlist.
+
+    Run once at ``safe_expr()`` call time -- the resulting Condition is only
+    built if validation passes, so rejected expressions never reach runtime.
+    """
+
+    def _reject(self, node: ast.AST, why: str) -> None:
+        raise ValueError(
+            f"safe_expr: disallowed construct in expression: {why} "
+            f"(at line {getattr(node, 'lineno', '?')}, col {getattr(node, 'col_offset', '?')})"
+        )
+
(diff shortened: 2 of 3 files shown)


skrawcz (commented):
-

skrawcz (approved):

PR #821, opened 2026-06-24
Add MongoDB driver handshake metadata

alexbevi:
## Summary

This PR identifies Burr in the MongoDB [driver handshake][handshake-spec], enabling server-side telemetry to attribute connections back to the library. MongoDB records this metadata at connection time, making it visible in`currentOp`, Atlas performance advisor, and `mongod` logs.

[handshake-spec]: https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.md

## What changes

A module-level constant is defined once in `b_pymongo.py`:

```python
_DRIVER_INFO = DriverInfo(name="Burr", version=_VERSION)
```

This is injected at every `MongoClient` construction site and, for the case where a caller passes in an existing client, via `append_metadata`.

## How it appears in the logs

On a MongoDB 4.4+ server, the `isMaster`/`hello` handshake carries the
following client metadata document (truncated to key fields):

```json
{
  "driver": {
    "name": "PyMongo|Burr",
    "version": "4.9.2|0.42.0"
  },
  "platform": "CPython 3.10.3.final.0"
}
```

In `mongod` diagnostic logs this appears as:

> `{"t":{"$date":"..."},"s":"I","c":"NETWORK","id":51800,"ctx":"conn1","msg":"client metadata","attr":{"remote":"127.0.0.1:54321","client":"conn1","doc":{"driver":{"name":"PyMongo|Burr","version":"4.9.2|0.42.0"},...}}}`

The `PyMongo|Burr` driver name lets operators distinguish Burr connections from bare PyMongo connections in server logs, Atlas metrics, and `db.currentOp()`.

## Testing

Six unit tests added in `tests/integrations/persisters/test_b_pymongo_driver_info.py`
(no live database required):

- `_DRIVER_INFO` has `name="Burr"` and version matching the installed package
- `from_values()` passes `driver=_DRIVER_INFO` to `MongoClient`
- `from_values()` does not override a caller-supplied `driver`
- `__init__()` calls `append_metadata(_DRIVER_INFO)` when the client supports it
- `__init__()` does not raise when the client lacks `append_metadata` (older pymongo)

No version constraint change is needed — `DriverInfo` is available since
pymongo 4.0, and `append_metadata` since 4.14; the existing unconstrained
`"pymongo"` dependency satisfies both.

## Checklist

### DIFF
--- a/burr/integrations/persisters/b_mongodb.py
+++ b/burr/integrations/persisters/b_mongodb.py
@@ -21,6 +21,9 @@ import logging
 
 from pymongo import MongoClient
 
+from burr.integrations.persisters.b_pymongo import (
+    _DRIVER_INFO,
+)
 from burr.integrations.persisters.b_pymongo import MongoDBBasePersister as PymongoPersister
 
 logger = logging.getLogger(__name__)
@@ -50,6 +53,7 @@ class MongoDBBasePersister(PymongoPersister):
 
         if mongo_client_kwargs is None:
             mongo_client_kwargs = {}
+        mongo_client_kwargs.setdefault("driver", _DRIVER_INFO)
         client = MongoClient(uri, **mongo_client_kwargs)
         return PymongoPersister(
             client=client,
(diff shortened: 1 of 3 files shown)


skrawcz (approved):
thanks!

Issue #400, opened 2024-10-16
Typing State for the class-based API does not work


mdrideout:
Following [docs](https://burr.dagworks.io/pull/370/concepts/state-typing/#action-level-typing) for action level typing for class-based actions does not work.

ref: https://github.com/DAGWorks-Inc/burr/issues/386

# Current behavior

**Example first action:**

```python
class SetInitialPromptAction(Action):
    @property
    def reads(self) -> list[str]:
        return []

    def run(self, state: ApplicationState, prompt: str) -> dict:
        return {"initial_prompt": prompt}

    @property
    def writes(self) -> list[str]:
        return ["initial_prompt"]

    def update(self, result: dict, state: ApplicationState) -> ApplicationState:
        prompt = result["initial_prompt"]
        logger.info(f"Saving prompt to state: {prompt}")
        state.initial_prompt = prompt
        return state

    @property
    def inputs(self) -> list[str]:
        return ["prompt"]
```

**Example second action:**

```python
class ExtractSetAction(Action):
    @property
    def reads(self) -> list[str]:
        return ["initial_prompt"]

    def run(self, state: ApplicationState) -> dict:
        logger.info(f"ApplicationState: {state}")

        # Read prompt from state
        prompt = state.initial_prompt
        ...
```

Logs: `ApplicationState: {'initial_prompt': None}`

## Stack Traces

<details>

api   | ********************************************************************************
api   | -------------------------------------------------------------------
api   | Oh no an error! Need help with Burr?
api   | Join our discord and ask for help! https://discord.gg/4FxBMyzW5n
api   | -------------------------------------------------------------------
api   | > Action: `extract_set` encountered an error!<
api   | > State (at time of action):
api   | {'__PRIOR_STEP': 'set_prompt',
api   |  '__SEQUENCE_ID': 1,
api   |  'initial_prompt': None,
api   |  'set_from_prompt': None}
api   | > Inputs (at time of action):
api   | {'prompt': 'bicep curls with 22 pound dumbells for 21 reps'}
api   | ********************************************************************************
api   | Traceback (most recent call last):
api   |   File "/usr/local/lib/python3.11/site-packages/burr/core/application.py", line 816, in _step
api   |     result = _run_function(
api   |              ^^^^^^^^^^^^^^
api   |   File "/usr/local/lib/python3.11/site-packages/burr/core/application.py", line 123, in _run_function
api   |     result = function.run(state_to_use, **inputs)
[...]

PR #823, opened 2026-06-27
ci: base stale-assignment clock on human activity, not updated_at

andreahlert:
## Problem

The `Stale Assignments` workflow (`.github/workflows/stale-assignments.yml`) is supposed to warn at 14 days of inactivity and unassign at 21. It **never unassigns**.

It measures inactivity from `issue.updated_at`:

```js
const updatedAt = new Date(issue.updated_at);
const daysSinceUpdate = Math.floor((now - updatedAt) / (1000 * 60 * 60 * 24));
```

`updated_at` is bumped by **any** change to the issue, including the bot's own warning comment and label edits. So each weekly warn resets the clock, capping `daysSinceUpdate` at the warn interval. The 21-day unassign branch is effectively dead code: the bot warns forever and never removes an assignee.

### Evidence (#400)

Bot warning timestamps on #400:

| Date | Text |
|---|---|
| 2026-04-15 | inactive for **17 days** |
| 2026-04-29 | inactive for **14 days** |
| 2026-06-03 | inactive for **16 days** |
| 2026-06-24 | inactive for **20 days** |

Apr 15 to Apr 29 is exactly 14 days. An honest counter would have read 31, not 14, the warning comment reset `updated_at`. The assignee was never removed despite four warnings.

## Fix

Measure inactivity from the last **human** activity, taking the most recent of:

- last non-bot comment (`github-actions[bot]` excluded),
- the `assigned` timeline event for a current assignee (so a freshly assigned issue is not instantly stale),
- issue creation time (fallback).

Bot comments no longer pollute the signal, so the clock accumulates past 14 and the 21-day unassign threshold becomes reachable. The paginated comments are reused for the existing recent-warn spam guard.

## Notes

- No behavior change to thresholds (14 warn / 21 unassign) or to the `lifecycle/frozen` exemption.
- Security note in the header still holds: only numeric IDs and login names from the API are used; no untrusted input reaches a shell.
- Validated locally: YAML parses, embedded script passes `node --check`.

### DIFF
--- a/.github/workflows/stale-assignments.yml
+++ b/.github/workflows/stale-assignments.yml
@@ -24,6 +24,10 @@
 #   - 21 days total -> unassign + add help wanted
 #   - Issues with lifecycle/frozen are exempt
 #
+# "Activity" = last human comment or the assignment event -- deliberately NOT
+# issue.updated_at, which is bumped by the bot's own comment and label edits and
+# would reset the clock every run (so the unassign threshold was never reached).
+#
 # Security: only uses numeric IDs and login names from the GitHub API.
 # No untrusted string input is interpolated into shell commands.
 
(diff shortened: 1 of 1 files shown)


andreahlert:
Superseding this. Rather than chase a correct activity signal, removing the stale-assignment bot entirely (it posted misleading inactivity numbers and never actually unassigned anyone). New PR removes the workflow.

andreahlert:
Reopened. Disregard the previous close note about removing the workflow, that was a misread. The bot stays; this PR is the intended fix: the bot's own warning comment no longer resets the inactivity clock. Excluding `github-actions[bot]` comments from the activity signal is the whole point, so warnings accumulate and the 21-day unassign threshold becomes reachable.

elijahbenizzy (approved):

PR #828, opened 2026-06-30
fix(tracking): S3 tracking server fails on first start with 'no such table: project' (#574)

### LINKED ISSUE #574: Burr S3 Tracking - Deployment and Docker Image
Hey Guys, 

I am using the Dockerfile provided to host s3 backed server tracker on EKS. I am able to get the UI and stuff and link S3 bucket but for some reason I run into the error 

`
[INFO] 2025-08-26 06:27:39,817 burr.tracking.server.run(79): Updating backend index...
[ERROR] 2025-08-26 06:27:39,819 burr.tracking.server.run(99): Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/tortoise/backends/sqlite/client.py", line 43, in translate_exceptions_
    return await func(self, query, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/tortoise/backends/sqlite/client.py", line 155, in execute_query
    rows = await connection.execute_fetchall(query, values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/aiosqlite/core.py", line 202, in execute_fetchall
    return await self._execute(self._execute_fetchall, sql, parameters)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/aiosqlite/core.py", line 122, in _execute
    return await future
           ^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/aiosqlite/core.py", line 105, in run
    result = function()
             ^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/aiosqlite/core.py", line 83, in _execute_fetchall
    cursor = self._conn.execute(sql, parameters)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sqlite3.OperationalError: no such table: project
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/fastapi_utils/tasks.py", line 90, in loop
    await _handle_func(func)
  File "/usr/local/lib/python3.11/site-packages/fastapi_utils/tasks.py", line 23, in _handle_func
    await func()
  File "/usr/local/lib/python3.11/site-packages/burr/tracking/server/run.py", line 80, in sync_index
    await backend.update()
  File "/usr/local/lib/python3.11/site-packages/burr/tracking/server/s3/backend.py", line 520, in update
    await self._update_projects()
  File "/usr/local/lib/python3.11/site-packages/burr/tracking/server/s3/backend.py", line 231, in _update_projects
    current_projects = await Project.all()
                       ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/tortoise/queryset.py", line 1235, in _execute
    instance_list = await self._db.executor_class(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/tortoise/backends/base/executor.py", line 108, in execute_select
    _, raw_results = await self.db.execute_query(sql, values)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/tortoise/backends/sqlite/client.py", line 45, in translate_exceptions_
    raise OperationalError(exc)
tortoise.exceptions.OperationalError: no such table: project
`

i am not able to grasp it, I am populating the S3 with the burr data itself from a client based on the architecture

vaquarkhan:
Root Cause
Bug 1 — no such table: project:

On a fresh container/EKS deploy with no pre-existing snapshot in S3, the tracking server crashes immediately when the indexer runs:

sqlite3.OperationalError: no such table: project
RegisterTortoise(app, config=..., add_exception_handlers=True) does not generate schemas by default (generate_schemas defaults to False). On first start there is no snapshot DB to download, so the SQLite file is created empty (no tables). When sync_index() calls backend.update() → _update_projects() → Project.all(), it hits a nonexistent table.

Bug 2 — indexer silently drops old logs:

The max_paths batch cap in _gather_paths_to_update had a broken break that only exited the inner for loop over a single S3 page, not the outer paginator loop. The paginator continued fetching additional pages, collecting far more than max_paths files. The watermark then advanced to the last file in this oversized batch, permanently skipping files that fell between position max_paths and the actual end of the batch on subsequent cycles.

Fix
Bug 1: Call Tortoise.generate_schemas(safe=True) inside the lifespan after RegisterTortoise enters. safe=True uses CREATE TABLE IF NOT EXISTS, so it:

Creates tables on first start (no snapshot)
Is a no-op when tables already exist from a downloaded snapshot
Never clobbers existing data
Bug 2: Add a cap_reached flag that breaks the outer paginator loop when the inner break fires. The batch is now truly capped at max_paths, and the watermark only advances to the last file in a correctly-sized batch.

Files Changed
backend.py
 — schema generation in lifespan (Bug 1); paginator break fix (Bug 2)
test_s3_backend_bug574.py
 — regression tests (moto-based)
Tests
test_generate_schemas_safe_true_creates_tables — verifies tables are created on empty DB
test_generate_schemas_safe_true_does_not_clobber_existing — verifies safe=True preserves snapshot data
test_gather_paths_respects_max_paths_cap — verifies batch cap works (requires moto)
test_watermark_advances_only_to_last_indexed_file — verifies watermark correctness (requires moto)
Bug 1 tests pass locally. Bug 2 tests require moto (marked skip if not installed; will run in CI).

Fixes #574

### DIFF
--- a/burr/tracking/server/s3/backend.py
+++ b/burr/tracking/server/s3/backend.py
@@ -36,7 +36,7 @@ from aiobotocore import session
 from fastapi import FastAPI
 from pydantic import field_validator
 from pydantic_settings import BaseSettings
-from tortoise import functions, transactions
+from tortoise import Tortoise, functions, transactions
 from tortoise.contrib.fastapi import RegisterTortoise
 from tortoise.expressions import Q
 
@@ -448,6 +448,7 @@ class SQLiteS3Backend(
         logger.info(f"Scanning log data for project: {project.name}")
         async with self._session.create_client("s3") as client:
             paginator = client.get_paginator("list_objects_v2")
+            cap_reached = False
             async for result in paginator.paginate(
                 Bucket=self._bucket,
                 Prefix=f"{self._data_prefix}/{project.name}/",
@@ -460,7 +461,10 @@ class SQLiteS3Backend(
                     logger.info(f"Found new file: {key}")
                     paths_to_update.append(DataFile.from_path(key, created_date=last_modified))
                     if len(paths_to_update) >= max_paths:
+                        cap_reached = True
                         break
+                if cap_reached:
+                    break
         logger.info(f"Found {len(paths_to_update)} new files to index")
         return paths_to_update
 
@@ -639,7 +643,15 @@ class SQLiteS3Backend(
     async def lifespan(self, app: FastAPI):
         if not os.path.exists(dirname := os.path.dirname(settings.DB_PATH)):
             os.mkdir(dirname)
-        async with RegisterTortoise(app, config=settings.TORTOISE_ORM, add_exception_handlers=True):
+        async with RegisterTortoise(
+            app,
+            config=settings.TORTOISE_ORM,
+            add_exception_handlers=True,
+        ):
+            # Ensure index tables exist on first start (no-op if tables already
+            # exist from a downloaded snapshot). safe=True uses CREATE TABLE IF
+            # NOT EXISTS, so it never clobbers existing data.
+            await Tortoise.generate_schemas(safe=True)
             yield
 
     async def list_projects(self, request: fastapi.Request) -> Sequence[schema.Project]:
--- a/tests/tracking/test_s3_backend_bug574.py
+++ b/tests/tracking/test_s3_backend_bug574.py
@@ -1,0 +1,258 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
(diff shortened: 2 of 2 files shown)


Haricharanpanjwani:
Thanks for putting this together @vaquarkhan. I think the runtime fix is headed in the right direction, especially the `generate_schemas(safe=True)` change for the fresh-start/no-snapshot case.

I do have two concerns around test coverage:

1. The new regression file does not appear to be exercised by CI. In `.github/workflows/python-package.yml`, the S3 tracking job still only runs `tests/tracking/test_bip0042_s3_buffering.py`, so `tests/tracking/test_s3_backend_bug574.py` is not currently part of the tracked workflow.

2. I don’t think the bug-2 tests actually reproduce the paginator bug from #574: `test_gather_paths_respects_max_paths_cap()` creates only 15 objects and does not force multi-page pagination. The original bug was that hitting `max_paths` only broke the inner loop, while the outer paginator loop kept fetching later pages. If everything comes back in one page, the pre-fix code would still pass this test.
   - `test_watermark_advances_only_to_last_indexed_file()` never exercises `_gather_paths_to_update()` or `_scan_and_update_db_for_project()`; it just computes `max(...)` over a handcrafted slice, so it doesn’t validate the actual watermark advancement behavior under the broken pagination path.

Would you mind tightening this by:
- wiring the new regression file into CI, and
- adding a test that forces multiple paginator pages and then verifies we stop after `max_paths` and persist the expected watermark?

Once that’s in place, I think this will be in much better shape.

PR #586, opened 2025-10-05
Various UI updates: remove elkjs add dagre, cost updates, executable script

skrawcz:
Various changes/

## Changes

* remove elkjs because of license and replace with dagre
* update model costs
* change script to executable for ergonomics.

## How I tested this
* spun up things locally

## Checklist

- [ ] PR has an informative and human-readable title (this will be pulled into the release notes)
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future TODOs are captured in comments
- [ ] Project documentation has been updated if adding/changing functionality.

PR #834, opened 2026-07-03
fix: tracking UI graph view freezes on state machines with dense transitions

### LINKED ISSUE #833: UI graph view unusable on large state machines: smart-edge pathfinding reruns for every edge on every render
I first hit this on one of my own Burr apps: clicking through steps in the tracking UI would freeze the tab for a second or more per click, and the bigger the state machine got, the worse it was. I dug in and it turns out the graph view re-runs A* pathfinding for every edge on every render.

# Current behavior

Clicking or hovering a step in the step list highlights the matching node and edges in the graph. On a small app that's instant. Past a few dozen actions it blocks the main thread for seconds per click.

I benchmarked it with Playwright against the production UI build, timing from click dispatch to the next paint, median of 12 clicks spread across the step list:

| Actions | Edges | Initial render | Click to paint (median) | Hover to paint (median) |
|---|---|---|---|---|
| 10 | 25 | 0.9s | 32ms | 30ms |
| 75 | 221 | 1.1s | 489ms | 495ms |
| 150 | 449 | 5.8s | 11,194ms (max 23,835ms) | 11,549ms |
| 300 | 899 | never renders, killed after 5 min | n/a | n/a |

Hover is just as expensive as click, and hover fires continuously while the pointer moves down the step list, so on a large app the graph pane is effectively frozen the whole time you're inspecting it.

A V8 profile during the click loop on the 150-action app shows 41% garbage collector, 16% `_buildNodes`, 12% `findPath`, plus `getNeighbors`. All of that is the A* pathfinding stack inside `@tisoap/react-flow-smart-edge` and the allocation churn it causes.

## Root cause

All in `telemetry/ui/src/components/routes/app/GraphView.tsx` (line numbers from current main):

1. `ActionActionEdge` calls `getSmartEdge` in its render body (line 162). That builds a pathfinding grid over the bounding box of all nodes and runs A* for that one edge. Nothing memoizes it, so it re-runs for every edge on every render. Per-edge cost grows with layout area, which grows with node count, so the total is superlinear in graph size.
2. Highlight state reaches nodes and edges through `NodeStateProvider`, and `_Graph` recreates the context value object inline on every render (lines 346-352). So every click and every hover enter/leave re-renders all N nodes and all E edges, and every edge re-runs A*.
3. `ActionActionEdge` also subscribes to the full node array via `useNodes()` (line 152), so any node store change re-renders every edge.
4. With auto-refresh on, the app refetches every 500ms (`REFRESH_INTERVAL`, `AppView.tsx:38`). react-query's structural sharing usually keeps the `data.application` reference stable across polls, but each poll still re-renders the view, and because of item 2 that re-renders every node and edge and re-runs A* for every edge, twice a second, with no interaction at all. Anything that does hand the graph a new object identity with the same structure (switching focus between a parent app and a structurally identical sub-app, for example) re-triggers the layout effect (`GraphView.tsx:332-343`) on top of that: full dagre relayout, new node/edge arrays, and a `fitView` that resets the viewport.

The 300-action hang is just item 1 paid once for all 899 edges over a very large grid at initial render.

1. Run the script below to log synthetic applications into `~/.burr`, then open the UI and open `ui-perf-bench/bench-150n`.
2. Click or hover step rows and feel the freeze. Open `bench-300n`: the graph never renders.

<details>
<summary>gen_large_app.py</summary>

```python
import argparse

from burr.core import ApplicationBuilder, State, action, default, expr


def step(state: State) -> State:
    return state.update(counter=state["counter"] + 1)


def build(n_nodes: int, extra_edges: int):
    names = [f"step_{i:03d}" for i in range(n_nodes)]
    transitions = []
    for i in range(n_nodes - 1):
        transitions.append((names[i], names[i + 1], default))
    for i in range(n_nodes):
        for k in range(1, extra_edges + 1):
            j = (i * 7 + k * 13 + 3) % n_nodes
            if j != i and j != i + 1:
                transitions.append((names[i], names[j], expr("counter < 0")))
    return names, transitions


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--nodes", type=int, default=150)
    parser.add_argument("--extra-edges", type=int, default=2)
    args = parser.parse_args()

    names, transitions = build(args.nodes, args.extra_edges)
    app = (
        ApplicationBuilder()
        .with_actions(**{name: step for name in names})
        .with_transitions(*transitions)
        .with_state(counter=0)
        .with_entrypoint(names[0])
        .with_identifiers(app_id=f"bench-{args.nodes}n")
        .with_tracker("local", project="ui-perf-bench")
        .build()
    )
    app.run(halt_after=[names[-1]])
    print(f"done: {args.nodes} nodes, {len(transitions)} edges")


if __name__ == "__main__":
    main()
```

</details>

Reproduced on the apache-burr 0.40.2 release build (macOS, Chromium headless via Playwright). The per-edge `getSmartEdge` call in the edge render body is in both 0.40.2 and current main, so both are affected. The file isn't identical between the two though: #586 swapped the layout engine from elkjs to dagre after 0.40.2, and when I re-ran the same benchmark on a main source build it came out much slower than 0.40.2 (9.7s vs 0.5s per click at 75 actions, numbers in the first comment below). That fits the mechanism: the pathfinding grid covers the layout's bounding box, and dagre spreads the same graph over more area, so every edge's A* run costs more.

# Expected behavior

The graph should repaint within a frame or two of a click or hover regardless of graph size, and a 300-action graph should render.

# Additional context

Sketch of a fix, all local to `GraphView.tsx`:

1. Memoize the `getSmartEdge` result per edge on geometry and node positions. Selection and hover don't move nodes, so highlighting stops paying for pathfinding entirely.
2. Above some node count, skip `getSmartEdge` and take the existing `getBezierPath` fallback so big graphs render at all (the code already falls back to bezier when `getSmartEdge` returns null).
3. Memoize the `NodeStateProvider` value in `_Graph`.
4. Key the relayout effect on graph structure (action names and transitions) instead of object identity, so identity churn without a structural change (focus switches, responses that aren't reference-stable) stops triggering full relayouts.

Happy to submit a PR.

msradam:
The tracking UI graph view freezes on state machines with a lot of transitions: every click or hover on a step re-runs A* pathfinding for every edge. I first hit this on one of my own apps (17 actions, 257 conditional transitions, about a one second freeze per click or hover). On current main a 75-action synthetic graph blocks ~10s per click and a 300-action graph never finishes its initial render. Benchmarks, profiles, and root cause are in #833.

Fixes #833

## Changes

All in `GraphView.tsx`, plus two call sites in `AppView.tsx`:

- Memoize the smart-edge path per edge, keyed on edge geometry and the node set. Highlighting doesn't move nodes, so clicks and hovers in the step list stop paying for pathfinding (interactions that change the node set itself still recompute, as before). This alone takes my real app from 999ms to 32ms per interaction, with rendering unchanged.
- Above 100 rendered nodes (`SMART_EDGE_NODE_LIMIT`), skip A* and use the existing `getBezierPath` fallback so very large graphs render at all. Below the threshold, edge routing is untouched.
- Memoize the `NodeStateProvider` context value instead of rebuilding it inline every render.
- Key the relayout effect on the rendered graph structure (node ids and types plus edge source/target/condition, derived from the same conversion the renderer uses, so the key can't disagree with what's drawn) instead of `stateMachine` object identity. react-query's structural sharing keeps the reference stable across most polls, so the practical per-poll cost was the context churn above, but anything that hands the graph a new identity with unchanged structure (switching focus between a parent app and a structurally identical sub-app, for example) used to trigger a full dagre relayout plus `fitView` and throw away the user's pan and zoom. Now relayout happens only when the graph actually changes.
- Build a fresh dagre graph per layout. The old module-level instance kept nodes from previously viewed applications.
- Remove a leftover `label={'test'}` debug prop, and pass `undefined` instead of a fresh `[]` literal for the currently unused `highlightedActions` prop (a new array identity each render would defeat the context memoization).
- Node data now carries only the rendered label instead of the whole `ActionModel`, so a replaced model object can't leave stale data behind in nodes.

## How I tested this

Playwright benchmark, click dispatch to next paint, median of 12 clicks, production builds of the same source where the only variable is this diff:

| Application | Unfixed | With fix |
|---|---|---|
| Synthetic, 75 actions / 221 edges | 9,669ms | 32ms |
| Synthetic, 150 actions / 449 edges | benchmark hung (>300s) | 31ms |
| Synthetic, 300 actions / 899 edges | never renders | 75ms |
| Real agent app, 17 actions / 257 transitions | 999ms | 32ms |
| Real agent apps at 9-13 nodes, and bundled `demo_chatbot` | 32ms | 32ms |

The 150 and 300 node synthetics are deliberate stress sizes to bound the improvement and find where rendering falls over; the 17-action agent app is the realistic case. Hover costs the same as click in every case. V8 profiles put the unfixed time in the smart-edge stack (`findPath`, `_buildNodes`, `getNeighbors`, plus the GC churn they cause); with the fix that stack is gone. The harness lives on [`spike/ui-graph-perf`](https://github.com/msradam/burr/tree/spike/ui-graph-perf/scripts/perf) if you want to re-run it.

For visual parity I screenshotted the graph pane on both builds: byte-identical PNGs (`cmp` exit 0) for the bundled `demo_chatbot` (below) and for a 75-node synthetic graph. Below the threshold, rendering is unchanged pixel for pixel. Above it the change is deliberate: bezier edges instead of A*-routed ones, in territory where the view previously took ~6s to appear on the 0.40.2 release (worse on this source) or didn't appear at all.

demo_chatbot graph pane, identical on both builds

I also checked the highlight behavior directly rather than trusting screenshots alone: clicking a step row applies the same classes to its graph node on both builds (`bg-green-500/80 ... border-dwlightblue/50 text-white border-2`), hovering a row applies `opacity-50` to its node, and full-page screenshots taken after 12 identical clicks and hovers are byte-identical across builds for four real applications.

After the second commit I re-ran the validation against the PR head: the dense real app benches 32ms clicks and 29ms hovers, the 150-action synthetic 32ms, the highlight and Show Inputs assertions pass, and the graph pane screenshot is still byte-identical to the unfixed build's.

I ran the CI checks locally the way the workflows define them: `npm run build`, `npm run lint:fix` (`--max-warnings=0`), and `npm run format:fix` all pass; pre-commit hooks pass on the changed files; `python -m pytest tests --ignore=tests/integrations/persisters --ignore=tests/integrations/test_bip0042_bedrock.py` gives 538 passed and 1 skipped, plus one Ray-startup timeout on my Mac that fails identically without this diff (there is no Python in it). `npm test` fails on main either way (`react-syntax-highlighter` ships ESM the CRA Jest config can't parse), which I assume is why `ui.yml` doesn't invoke it.

## Notes

- The 100-node threshold is a judgment call from the measurements: at 75 nodes smart edges still complete initial rendering (and output is unchanged by this diff), while at 150 nodes the unfixed benchmark on this source hung outright and the 0.40.2 release cost 11s per click. Cost grows with edge count times layout area, so a node-count cutoff is blunt but predictable. Happy to tune the constant or make it configurable.
- The unfixed numbers here are worse than the 0.40.2-based numbers in #833 because #586 swapped elkjs for dagre after 0.40.2, and the dagre layout covers more area, which the pathfinding grid scales with. Same mechanism and same fix on both versions.
- The threshold counts rendered nodes, which includes input nodes when Show Inputs is on, so toggling that checkbox can move a graph across the limit and switch edge styles. That tracks the real pathfinding cost (the grid covers everything rendered), but it's worth knowing it's mode-dependent.
- There's one `eslint-disable-next-line react-hooks/exhaustive-deps` on the relayout effect: depending on the structure key instead of the object it reads is the point of that change. (Under the current lint config the rule is off, so the comment is documentation for whenever the config tightens.)
- Initial render of very large graphs is still slow (~60s at 300 actions, dominated by dagre layout and first paint of 899 edges). That's a separate follow-up, noted in #833.

## Checklist

### DIFF
--- a/telemetry/ui/src/components/routes/app/AppView.tsx
+++ b/telemetry/ui/src/components/routes/app/AppView.tsx
@@ -542,7 +542,7 @@ export const AppView = (props: {
                   stateMachine={data.application}
                   currentAction={currentStep}
                   // highlightedActions={previousActions}
-                  highlightedActions={[]}
+                  highlightedActions={undefined}
                   hoverAction={hoverAction}
                 />
               </div>
@@ -555,7 +555,7 @@ export const AppView = (props: {
             stateMachine={currentFocusStepsData?.application || data.application}
             // stateMachine={data.application}
             // highlightedActions={previousActions}
-            highlightedActions={[]}
+            highlightedActions={undefined}
             hoverAction={hoverAction}
             currentActionLocation={currentSequenceLocation}
             displayGraphAsTab={displayGraphAsTabs} // in this case we want the graph as a tab
--- a/telemetry/ui/src/components/routes/app/GraphView.tsx
+++ b/telemetry/ui/src/components/routes/app/GraphView.tsx
@@ -20,7 +20,7 @@
 import { ActionModel, ApplicationModel, Step } from '../../../api';
 
 import dagre from 'dagre';
-import React, { createContext, useLayoutEffect, useRef, useState } from 'react';
+import React, { createContext, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import ReactFlow, {
   BaseEdge,
   Controls,
@@ -38,8 +38,6 @@ import 'reactflow/dist/style.css';
 import { backgroundColorsForIndex } from './AppView';
 import { getActionStatus } from '../../../utils';
 import { getSmartEdge } from '@tisoap/react-flow-smart-edge';
-
-const dagreGraph = new dagre.graphlib.Graph();
 
 const dagreOptions = {
   rankdir: 'TB', // Top to bottom layout (equivalent to ELK's UP direction)
@@ -138,6 +136,10 @@ const InputNode = (props: { data: NodeData }) => {
     </>
   );
 };
+// Past this size, smart-edge A* pathfinding is too slow to run at all -- a 300-node graph
+// never finishes its initial render. See https://github.com/apache/burr/issues/833
+const SMART_EDGE_NODE_LIMIT = 100;
+
 // TODO -- separate out into different edge types
 export const ActionActionEdge = ({
   sourceX,
@@ -159,20 +161,24 @@ export const ActionActionEdge = ({
   );
   const containsTo = allActionsInPath.some((action) => action.step_start_log.action === data.to);
   const shouldHighlight = containsFrom && containsTo;
-  const getSmartEdgeResponse = getSmartEdge({
-    sourcePosition,
-    targetPosition,
-    sourceX,
-    sourceY,
-    targetX,
-    targetY,
-    nodes
-  });
-  let edgePath = null;
-  if (getSmartEdgeResponse !== null) {
-    edgePath = getSmartEdgeResponse.svgPathString;
-  } else {
-    edgePath = getBezierPath({
+  // Memoized: highlight changes re-render every edge, and pathfinding must not rerun then.
+  // See https://github.com/apache/burr/issues/833
+  const edgePath = useMemo(() => {
+    if (nodes.length <= SMART_EDGE_NODE_LIMIT) {
+      const getSmartEdgeResponse = getSmartEdge({
+        sourcePosition,
+        targetPosition,
+        sourceX,
+        sourceY,
+        targetX,
+        targetY,
+        nodes
+      });
+      if (getSmartEdgeResponse !== null) {
+        return getSmartEdgeResponse.svgPathString;
+      }
+    }
+    return getBezierPath({
       sourceX,
       sourceY,
       sourcePosition,
@@ -180,7 +186,7 @@ export const ActionActionEdge = ({
       targetY,
       targetPosition
     })[0];
-  }
+  }, [nodes, sourceX, sourceY, targetX, targetY, sourcePosition, targetPosition]);
 
   const style = {
     markerColor: shouldHighlight ? 'black' : 'gray',
@@ -188,7 +194,7 @@ export const ActionActionEdge = ({
   };
   return (
     <>
-      <BaseEdge path={edgePath} markerEnd={markerEnd} style={style} label={'test'} />
+      <BaseEdge path={edgePath} markerEnd={markerEnd} style={style} />
     </>
   );
 };
@@ -201,7 +207,8 @@ const getLayoutedElements = (
   const isHorizontal = options?.['direction'] === 'LR';
   const direction = isHorizontal ? 'LR' : 'TB';
 
-  // Configure dagre graph
+  // Fresh graph per layout -- a shared one accumulates stale nodes across applications
+  const dagreGraph = new dagre.graphlib.Graph();
   dagreGraph.setDefaultEdgeLabel(() => ({}));
   dagreGraph.setGraph({
     ...dagreOptions,
@@ -329,6 +336,18 @@ export const _Graph = (props: {
 
   const { fitView } = useReactFlow();
 
+  // Keyed on structure rather than object identity: polling refetches produce a new
+  // stateMachine object every 500ms, which must not trigger a full relayout.
+  // See https://github.com/apache/burr/issues/833
+  const structureKey = useMemo(
+    () =>
+      JSON.stringify([
+        props.stateMachine.actions.map((action) => [action.name, action.inputs]),
+        props.stateMachine.transitions.map((t) => [t.from_, t.to, t.condition])
+      ]),
+    [props.stateMachine]
+  );
+
   useLayoutEffect(() => {
     const [nextNodes, nextEdges] = convertApplicationToGraph(props.stateMachine, showInputs);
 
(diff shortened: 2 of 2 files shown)


skrawcz (approved):
Really strong, well-motivated fix — the root cause (A* re-running per edge on every highlight, plus a shared module-level dagre graph and relayout keyed on object identity) is correctly diagnosed, and the memoization + structure-key + fresh-graph-per-layout changes address it cleanly. Scope is tight (two UI files), backwards compatibility is a non-issue (no Python/API/serialization touched; the reshaped node `data` type has no other consumers), and the benchmark/visual-parity writeup is exemplary.
One thing before merge: there's no automated coverage added and `npm test` is broken on `main`, so the "change in functionality is tested" checklist item is only satisfied manually via the out-of-tree Playwright harness; please either land a minimal test, commit/link the harness, or note in the PR that validation is manual so it's an explicit decision.
Minor: `convertApplicationToGraph` now runs twice per render that changes the structure key (once for the key, once in the effect) plus a `JSON.stringify` — measured-fine, but could be computed once and shared. Nice catches on the stale-dagre-graph and pan/zoom-preservation bugs.

---

- **File:** `telemetry/ui/src/components/routes/app/GraphView.tsx` — **Line ~333** (`const structureKey = useMemo(() => { const [nextNodes, nextEdges] = convertApplicationToGraph(...) ...`)
  **[SUGGESTION]** `convertApplicationToGraph` is now also called inside the layout effect below, so it runs twice per structure change, plus a `JSON.stringify` of the full node/edge arrays. Fine at the measured sizes, but you could compute the graph once (e.g. a single `useMemo` returning `{ nodes, edges, key }`) and feed both the effect and the key from it, avoiding the double conversion and any chance of the key disagreeing with what's laid out.
- **File:** `telemetry/ui/src/components/routes/app/GraphView.tsx` — **Line ~360** (`// eslint-disable-next-line react-hooks/exhaustive-deps`)
  **[NIT]** Good that this is documented. Since `props.stateMachine` and `showInputs` are the only inputs `structureKey` derives from, consider a short inline note that `fitView` is the only "real" extra dep and everything else is intentionally funneled through `structureKey`, so a future reader doesn't re-add `props.stateMachine` and reintroduce the identity-churn relayout.
- **File:** `telemetry/ui/src/components/routes/app/GraphView.tsx` — **Line ~180** (`}, [nodes, sourceX, sourceY, targetX, targetY, sourcePosition, targetPosition]);`)
  **[NIT]** Confirmed correct: highlight state comes from context, not node data, so `useNodes()` stays referentially stable on highlight and the memo actually saves the A* work. `shouldHighlight`/`style` stay outside the memo so visuals still update. No change needed — noting for the record.
- **File:** `telemetry/ui/src/components/routes/app/AppView.tsx` — **Line 545 & 558** (`highlightedActions={undefined}`)
  **[SUGGESTION]** Correct and intentional (avoids a fresh `[]` identity defeating the context memo). Since `previousActions` is commented out just above both, consider deleting those dead `// highlightedActions={previousActions}` comment lines while you're here to leave the call sites clean.

PR #843, opened 2026-07-22
fix: use utf-8 in CI smoke server script

Ghraven:
## Problem
`scripts/ci_smoke_server.py` writes and reads text artifacts without specifying an encoding. On Windows or other non-UTF-8 default locales, the generated app script or captured server log can become locale-dependent even though the script content is text.

## Before / after
Before:
- The smoke server log was opened with the platform default text encoding.
- The failure log dump used `Path.read_text()` without an explicit encoding.
- The generated Burr app script used `Path.write_text()` without an explicit encoding.

After:
- All three text boundaries use `encoding="utf-8"`.
- Binary/network handling is unchanged.

## Verification
- `python -m py_compile scripts/ci_smoke_server.py`

### DIFF
--- a/scripts/ci_smoke_server.py
+++ b/scripts/ci_smoke_server.py
@@ -153,7 +153,7 @@ def main() -> None:
         env = os.environ.copy()
         env["burr_path"] = str(burr_data_dir)
         env["PYTHONUNBUFFERED"] = "1"
-        with open(server_log, "w") as log_fh:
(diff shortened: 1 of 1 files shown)


skrawcz (approved):

skrawcz (approved):

Ghraven:
Thank you for the review and merge, I appreciate it!

Issue #205, opened 2024-05-26
Add ability to inject global inputs `__foo` variables


elijahbenizzy:
**Is your feature request related to a problem? Please describe.**
There are some inputs that everyone might want. Tracers, etc... We should be able to inject it.

**Describe the solution you'd like**

Thinking something like this:

```python
(
    ApplicationBuilder()....with_global_inputs(
    __langfuse_tracer=(app_id, ...: LangFuse(...), NoOpLangfuse())) # other parameters/whatever you call for langfurse
    .build()
)
```

Or some other way to provide a no-op integrator.

**Describe alternatives you've considered**
Could do it with inputs manually.

andreahlert:
Labels migrated to new taxonomy (#692).

andreahlert:
Closing as stale. No discussion or external demand since it was opened. Can be reopened if this becomes a real need.

Issue #46, opened 2024-02-28
Add metadata capture


skrawcz:
You want to +1 or -1 a response for later evaluation. We need to expose a way to do that.

# idea

```python
# annotated current returned state 
app.annotate({"value": 1, }) # implicit app_id, sequence_id

# annotate some thing afterwards
Tracker().annotate(app_id, sequence_id, {"value": 1})
Tracker().annotate(app_id, sequence_id, {"value": 1, "target": SOME_STATE_KEY})

```
You'd have to provide more spcificis on what the metadata is yourself I think. e.g. what it targets in the state.

This should work for arbitrary metadata annotation.

We'd then want some export functionality.

elijahbenizzy:
Some thoughts out loud:

More can go through the app -- having it on the tracker is weird...

```python
app.annotate({"value": 1})
app.annotate({"value": 1, }, sequence_id=10)
app.annotate({"value": 1, }, sequence_id=-10) # 10 back from current
```

Does the `target` make sense? Or does that just tie to the value?

elijahbenizzy:
Types of metadata we want:

1. Artifacts -- these are associated with traces/whatnot. These are immutable, and only logged during execution. Examples:
    - An embedding of something
    - The cost of an OpenAI call
    - A metric associated with an epoch for evaluating loss
    - A set of metrics that we might want to query for
    - A dataset (training data, whatnot)
2. Annotations -- these are added after the fact to allow for post-hoc analysis. E.G.
    - A user marking a response as good/bad
    - A user saying they like a model/like the results
    - Marking a dataset as useful/interesting in some way

andreahlert:
Labels migrated to new taxonomy (#692).

andreahlert:
Closing as stale. The scope here is too broad and it hasn't moved in 2 years. If specific parts (artifacts, annotations, querying) become priority, they should be individual issues.

PR #844, opened 2026-07-22
feat: add Langfuse tracing integration (LangfuseBridge)

### LINKED ISSUE #206: Langfuse integration
We should be able to automatically track every Burr app using Langfuse automatically.

With hooks (https://burr.dagworks.io/concepts/hooks/) one should be able to do this nicely.

A trace-based langfuse integration.  The following features at a minimum:
- [ ] One trace per application run (this is dependent on this: https://github.com/DAGWorks-Inc/burr/issues/203)
- [ ] One span per step
- [ ] One span per tracer span call
- [ ] Observations at will (perhaps state elements)
- [ ] Perhaps access to the langfuse client.
 
You'll want to follow the same classes as this does: https://github.com/DAGWorks-Inc/burr/blob/main/burr/tracking/client.py#L60,
except for the `BaseStateLoader`, that's not relevant. Then you can store state in it. 

See this for an inspiration (this is specific to datadog): https://github.com/DAGWorks-Inc/hamilton/blob/96da3107666dcccdd80da39d7d4d118b4de82571/hamilton/plugins/h_ddog.py#L19.

There are two ways to do this:
1. Expose the langfuse client in the functions/classes
2. Bury it behind the logger

We should enable both, but strive towards (2).

For (1), we can do it if you pass the langfuse object in for application calls as runtime inputs. This will make it easier: https://github.com/DAGWorks-Inc/burr/issues/205. 

For (2) we'll need to reconcile our model with theirs. Specifically, this will help: https://github.com/DAGWorks-Inc/burr/issues/46.

Let's do (1) then open up another issue for (2)

**Additional context**

hannguyen0712:
Adds a built-in Langfuse integration (#206): LangfuseBridge, a lifecycle hook that logs Burr application execution to Langfuse, built on OpenTelemetryBridge and the OTel-native Langfuse Python SDK.

Closes #206

## Changes

1. New `burr/integrations/langfuse.py` with `LangfuseBridge`, attached via `.with_hooks(LangfuseBridge())`:
  -   One trace per application execution call, one span per step, one span per `__tracer` span
  -   Action inputs/read state and results/written state captured as observation input/output (`capture_state=False` to disable)
  -   `__tracer.log_attributes()` captured as observation metadata (complex values JSON-serialized, since OTel attributes only accept primitives)
  -   `app_id` maps to the Langfuse session, `partition_key` to the user (both overridable); the client is exposed as `bridge.langfuse_client`
  -   Langfuse SDK v4 only exports LLM-relevant spans by default, so the bridge injects a `should_export_span` filter (exported as `burr_span_export_filter` for users constructing their own client), with a fallback for v3

2. Docs page (`docs/reference/integrations/langfuse.rst`), runnable example (`examples/integrations/langfuse/`), and a `langfuse` pyproject extra

**Note on the divergence from the tracking-client pattern in the issue:** the original issue proposed mirroring `burr/tracking/client.py` and depended on #205 (closed as not planned). Since then, the Langfuse SDK was rebuilt on OpenTelemetry and registers its span processor on the global tracer provider, so building on the existing `OpenTelemetryBridge` avoids duplicating span-lifecycle machinery and nests third-party OTel LLM instrumentation (e.g. `opentelemetry-instrumentation-openai`) inside step spans for free. This direction was discussed and approved in the issue thread (May 31). Client access, which #205 would have provided via runtime inputs, is handled through constructor injection and the public `bridge.langfuse_client` attribute instead.

Each execution call (`run`, `step`, `iterate`, ...) becomes its own trace; repeated `.step()` calls therefore produce multiple traces, grouped under one Langfuse session via `app_id` (`session_id` is overridable if that's not desired). Async and streaming execution get the same coverage as the existing OTel bridge, since all execute methods flow through the same hooks.

## How I tested this

- 6 integration tests (`tests/integrations/test_burr_langfuse.py`) run a real two-action app with a tracer span and logged attributes against an in-memory OTel exporter and a mocked Langfuse client, so CI needs no credentials. They assert the full span tree, parenting, session/user attributes, observation input/output payloads, metadata serialization, `capture_state=False`, constructor validation, the export filter, and error status.
- Verified end to end against Langfuse Cloud with `opentelemetry-instrumentation-openai` installed: screenshots below show the trace tree with the OpenAI generation (tokens/cost) nested inside the step spans, observation input/output, the session/user mapping, and error status on a failed run.
- Full `tests/core`, `tests/visibility`, and existing OTel integration tests pass; pre-commit clean.

### Verification screenshots (Langfuse Cloud)

**Image 1: Langfuse trace list of one trace per execution call**

langfuse-trace-list

Each `.run()` call produces one Langfuse trace rooted at a `run` span, with step and tracer spans nested beneath it.

**Image 2: Nested OpenAI generation view. Third-party OTel instrumentation composes automatically**

langfuse-openai-query-token-view

With `opentelemetry-instrumentation-openai` installed, the LLM call appears as a Langfuse generation nested inside the `_query_openai` tracer span, with model, role-tagged prompt/completion messages, token usage (334 → 322, ∑ 656), and cost. No Langfuse-specific code in the action.

**Image 3: When a run is invalid, error status displays properly**

   

A run with an invalid OpenAI key: the `chat_response` step span and the generation are marked ERROR with the full 401 message, and Langfuse flags the trace via the child errors. The root `run` span itself carries no error status, and this is the pre-existing core gap described in section Notes below (`post_run_execute_call` receives `exception=None`), not a limitation of this integration; the test comment marks the assertion to strengthen once that's fixed. Metadata also shows `burr.sequence_id` and the `burr.integrations.langfuse` tracer scope.

**Image 4: Tracer spans + logged attributes as observation metadata**

   

The `prepare_history` span opened via `__tracer(...)` appears under its step, and `__tracer.log_attributes(history_length=...)` lands in the span's Langfuse metadata.

**Image 5: Step state captured as observation input/output**

langfuse-chat-response-input-output

The `chat_response` step's read state (`prompt`, `chat_history`) shows as observation input and its written state as observation output (disable with `capture_state=False`).

**Image 6: Langfuse session view indicates that user mapping works correctly**

   

Burr's `app_id` maps to the Langfuse session (both `.run()` calls of one application grouped together) and `partition_key` to the user (`example-user`), enabling Langfuse's session and user views.

## Notes

- **Pre-existing core gap found while testing:** `post_run_execute_call` hooks always receive `exception=None`; the wrappers in `_call_execute_method_pre_post` (`burr/core/application.py`) initialize `exc = None` inside a `try/finally` with no `except`, so the exception is never captured. This affects the existing `OpenTelemetryBridge` too: root spans of failed runs are never marked errored (Langfuse still flags the trace via child-observation errors). This PR deliberately does not touch core; the error test asserts on the step span only, with a comment marking the assertion to strengthen once fixed. I have an 8-line fix with sync/async regression tests ready and am happy to file an issue and/or separate PR, whichever is preferred.
- The original issue said to open a follow-up for approach (2); I can file a tracking issue for the deferred items (deeper model reconciliation per #46, Langfuse-native generation types, scoring helpers) if useful.

## Checklist

- [X] Changes are limited to a single goal (no scope creep)
- [X] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.
- [X] Any _change_ in functionality is tested
- [X] New functions are documented (with a description, list of inputs, and expected output)
- [X] Placeholder code is flagged / future TODOs are captured in comments
- [X] Project documentation has been updated if adding/changing functionality.


jernejfrank (approved):
Looks good, awesome!

PR #852, opened 2026-07-25
Use UTF-8 for adaptive CRAG docs

Ghraven:
## Summary
- read adaptive CRAG example text documents with explicit UTF-8 encoding
- keep the existing short-document merge behavior unchanged

## Problem
`Path.read_text()` uses the platform default encoding when no encoding is provided. On Windows or other non-UTF-8 locales, the adaptive CRAG example can fail when docs contain non-ASCII content.

## Before / After
Before: docs were decoded with the local default encoding.

After: docs are decoded as UTF-8 consistently.

## Verification
- `python -m py_compile examples/adaptive-crag/application.py`

### DIFF
--- a/examples/adaptive-crag/application.py
+++ b/examples/adaptive-crag/application.py
@@ -101,8 +101,9 @@ def load_burr_docs(
         files_iter = file_names
     docs = []
     for file in files_iter:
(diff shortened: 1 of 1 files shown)


jernejfrank (approved):
Looks good

PR #853, opened 2026-07-26
fix(release): surface bundled JavaScript licenses

### LINKED ISSUE #756: License: surface bundled JS dependencies in wheel's LICENSE rather than only the webpack-generated LICENSE.txt
Raised by Jarek Potiuk during the 0.42.0-incubating RC3 PPMC review. Non-blocking for the release.

## Issue

The wheel ships the built React UI under `burr/tracking/server/build/`. That bundle (~1.7 MB total) embeds many third-party npm packages (React, react-router, lodash, the Magic UI runtime deps, etc.). Their license terms are documented in `burr/tracking/server/build/static/js/main.<hash>.js.LICENSE.txt` — which is webpack's auto-generated comment-banner aggregation.

That file is technically present and machine-readable, but it's not surfaced anywhere a typical license auditor would look: the wheel's top-level `LICENSE-wheel` does not reference it, and there's no `dist-info/licenses/` entry for the bundled deps.

## What's expected for ASF wheels

Apache release policy expects all bundled dependencies to be enumerated in the project's LICENSE / NOTICE in some form, even if the underlying license file is shipped separately. The current state passes RAT (we exclude `*.js`), but is fragile to a strict licensing review.

## Suggested fix

One of:
1. Have the release script copy `burr/tracking/server/build/static/js/main.*.js.LICENSE.txt` into `dist-info/licenses/LICENSE-bundled-js.txt` during wheel build, AND add a one-line pointer in `LICENSE-wheel` pointing readers to that file.
2. Pre-process the webpack output to extract a deduplicated, sorted, top-level `LICENSE-bundled-js.txt` that's appended to `LICENSE-wheel` directly.
3. Vendor a static `LICENSE-bundled-js.txt` listing the npm deps + their licenses, and verify it stays in sync via CI.

Option 1 is the lowest-effort and survives webpack hash changes automatically.

## References

- PPMC vote thread: https://www.mail-archive.com/dev@burr.apache.org/msg00127.html
- The auto-generated file in question: `burr/tracking/server/build/static/js/main.<hash>.js.LICENSE.txt` (present in every wheel build)

bm1016bm-svg:
Surfaces webpack-generated third-party JavaScript notices in the wheel's standard license directory, so consumers do not have to discover them inside package data.

## Changes

- include `main.*.js.LICENSE.txt` through PEP 639 `license-files`
- point wheel consumers to the bundled notices from `LICENSE-wheel`
- fail release artifact verification when the notices are absent

## How I tested this

- `python -m pytest tests/test_verify_apache_artifacts.py tests/test_release_config.py` (20 passed)
- built a wheel with Flit and inspected it to confirm the notice appears under `.dist-info/licenses/burr/tracking/server/build/static/js/`

## Notes

Closes #756.

## Checklist

- [x] Relevant checks and tests pass
- [x] The functionality change is tested
- [x] New helper behavior is self-describing and covered by tests
- [x] No placeholder code was added
- [x] License documentation was updated

### DIFF
--- a/LICENSE-bundled-js
+++ b/LICENSE-bundled-js
@@ -1,0 +1,110 @@
+/**
+ * @license React
+ * react-dom.production.min.js
+ *
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * @license React
+ * react-jsx-runtime.production.min.js
+ *
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * @license React
+ * react.production.min.js
+ *
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * @license React
+ * scheduler.production.min.js
+ *
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * @license React
+ * use-sync-external-store-shim.production.min.js
+ *
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * @license React
+ * use-sync-external-store-shim/with-selector.production.min.js
+ *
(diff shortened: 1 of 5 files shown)


jernejfrank:
This makes the entire CI matrix fail before tests run. The new license-
  files glob points into the gitignored burr/tracking/server/build
  directory, which is absent in a clean checkout, so Flit errors with No
  files found for [project.license-files]. Could we store the license
  notice at a stable, committed path instead?

bm1016bm-svg:
Addressed: the bundled UI notice now lives at the committed root-level LICENSE-bundled-js path, and both wheel metadata and release validation use that stable file. The focused suite passes (18 tests), and a wheel build includes all four expected license files. The new Apache Actions runs currently require maintainer approval before jobs can start.

bm1016bm-svg:
Closing as superseded by #868, which solves #756 more comprehensively by generating `LICENSE-wheel` from the built UI dependency graph and enforcing it in release validation. Thanks for the review and for catching the clean-checkout packaging issue.

PR #869, opened 2026-08-03
Fix the shipped npm advisories, and actually run the UI test suite

potiuk:
Stacked on #868 — merge that first; this branch is cut from it, so the
diff here will show its 4 commits until it lands.

Picks up the remaining non-blocking items from my 0.43.0 RC0 review, plus
two things found while fixing them.

### 1. npm advisories in shipped dependencies (16 → 2)

A fresh `npm ci` from the shipped lockfile reported 16 advisories, 7 of
them outside devDependencies. Checked each against the build sourcemap;
two reached users in the compiled bundle:

- **mdast-util-to-hast 13.1.0** — moderate, unsanitized `class` attribute.
  28 modules in the bundle, reached via `react-markdown`, which renders
  trace and chat content. Fixed by 13.2.1.
- **react-router 7.18.1** — high, RSC mode CSRF bypass.

`npm audit fix` (no `--force`) resolves everything semver-compatible. The
rest of the 16 (`postcss`, `minimatch`, `glob`, `brace-expansion`, `vite`)
are build-time only — confirmed 0 modules each in the shipped bundle.

**The 2 that remain are react-router / react-router-dom**, still inside
the advisory range at 7.18.2. The only fix is react-router 8.3.0, a major
bump. The advisory covers RSC mode, and the tracking UI is a static
client-side SPA with no RSC and no server actions, so it isn't reachable
here. I'd rather that upgrade be done on its own merits than smuggled in
as a security fix — happy to open it separately if you want it now.

`LICENSE-wheel` is regenerated for the two bundled version changes. Worth
noting the `--check` gate from #868 caught this drift on its first real
outing, which is exactly what it's for.

### 2. `useTheme` could take down the whole app

`useTheme` read and wrote `window.localStorage` directly, guarded only by
`typeof window === 'undefined'`. That assumption doesn't hold: a sandboxed
iframe raises `SecurityError` on access, browsers with storage disabled
omit it, and Node's own experimental `localStorage` global shadows jsdom's
under vitest. In all of those the hook throws during the initial
`useState` — which fails the entire `<App>` render, not just the theme.

Reads and writes now go through a helper that returns `null` when storage
is unreachable. The theme falls back to the system preference and just
isn't persisted.

### 3. The UI test suite was never run in CI

`ui.yml` builds, lints and formats, but never invokes vitest — so
`App.test.tsx`, the only UI test, ran nowhere. It has been failing on
newer Node with no signal at all. Added `npx vitest run` (plain `npm test`
starts watch mode and would hang the job).

Unrelated but worth a look some time: `ui.yml` runs `lint:fix` and
`format:fix` in CI, which *modify* files rather than check them, so lint
and format drift can't fail a build.

### Verification

- `npm audit --omit=dev`: 7 → 2, both react-router
- `npx vitest run`: 1 passed (was 1 failed)
- `npm run build`: clean, tsc included
- `generate_wheel_license.py --check`: up to date after regeneration
- `pytest tests/`: 687 passed, 4 skipped

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -46,5 +46,7 @@ jobs:
         node-version: ${{ matrix.node-version }}
     - run: npm install --ignore-scripts
     - run: npm run build
+    # `npm test` starts vitest in watch mode, which never exits on CI.
+    - run: npx vitest run
     - run: npm run lint:fix
     - run: npm run format:fix
--- a/LICENSE-wheel
+++ b/LICENSE-wheel
@@ -490,7 +490,7 @@ The following components are licensed under the MIT license:
   mdast-util-phrasing@4.1.0
       Copyright (c) 2017 Titus Wormer <tituswormer@gmail.com>
       Copyright (c) 2017 Victor Felder <victor@draft.li>
-  mdast-util-to-hast@13.1.0
+  mdast-util-to-hast@13.2.1
       Copyright (c) 2016 Titus Wormer <tituswormer@gmail.com>
   mdast-util-to-markdown@2.1.0
       Copyright (c) 2020 Titus Wormer <tituswormer@gmail.com>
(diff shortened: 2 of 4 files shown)


skrawcz:
@potiuk mind rebasing please / fixing the conflict?

potiuk:
sure :)

jernejfrank (approved):
cool, thanks!

PR #882, opened 2026-08-13
fix(release): address 0.43.0 RC feedback

jernejfrank:
## Summary

Address the non-blocking follow-ups reported while verifying 0.43.0 RC1:

- install the Graphviz Python package through the `tests` extra
- derive the official build epoch from the anchored Git `HEAD` timestamp
- recover and propagate that same epoch from source and sdist archive members for voter/CI rebuilds
- retain byte-for-byte SHA-512 verification of rebuilt sdists
- update the tracking UI transitive `nanoid` dependency from 3.3.16 to 3.3.18
- make UI CI use a clean lockfile install and non-mutating ESLint/Prettier checks
- fix the existing UI formatting findings and enforce zero lint warnings

## Validation

- release/verifier/Graphviz tests — 39 passed
- Python 3.9 syntax check
- full isolated release plus voter rebuild — rebuilt sdist SHA-512 and wheel contents matched
- extracted-sdist wheel build without `.git` — passed
- clean `npm ci`
- `npm audit` — zero vulnerabilities
- `npm run build`
- `npx vitest run` — 1 passed
- `npm run lint`
- `npm run format`
- `LICENSE-wheel` covers all 130 bundled npm packages
- focused pre-commit hooks — passed

### DIFF
--- a/.github/workflows/release-validation.yml
+++ b/.github/workflows/release-validation.yml
@@ -297,7 +297,11 @@ jobs:
           BURR_VERSION: ${{ needs.build-artifacts.outputs.version }}
         run: |
           mkdir -p /tmp/sdist-extract /tmp/sdist-wheel
-          tar -xzf "dist/apache-burr-${BURR_VERSION}-incubating-sdist.tar.gz" \
+          SDIST_PATH="dist/apache-burr-${BURR_VERSION}-incubating-sdist.tar.gz"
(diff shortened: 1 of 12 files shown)


elijahbenizzy (approved):
Reviewed the conflict resolution against #899 and re-ran tests, pre-commit, and the UI lint/format gate locally. LGTM.

PR #893, opened 2026-08-18
Document how to hand off control to a human

### LINKED ISSUE #56: Document how to hand-off to a human
If someone is building an agent, an agent can go do stuff, but then might need help. 

Frame that example using `run(halt_before=["human_step"])`...

TODO:

- [ ] write up documentation to explicitly explain this.

Excelius-Wang:
Closes #56.

Handing control back to a person is a pattern Burr already supports well, but the mechanics were only discoverable by reading `examples/email-assistant` and piecing together `halt_before` from the applications page. This adds a concepts page that states the pattern directly.

## Changes

- New page `docs/concepts/human-in-the-loop.rst`, listed in `docs/concepts/index.rst` after `state-machine`.
- Cross-references to it from the `Runtime Inputs` section of `actions.rst` and the `Step` section of `state-machine.rst`, since those are where a reader is most likely to be when the question comes up.

The page covers: why the person stays outside the graph; the `run(halt_before=[...])` / collect / `run(inputs=...)` loop and the fact that `halt_before` takes precedence when an action is in both halt lists; a minimal graph with a single approval step; a chat loop where the human action is both entrypoint and halt target; multi-handoff applications via `examples/email-assistant`; what changes in a web server; and why `input()` inside an action is CLI-only.

## How I tested this

- Ran every code block on the page against the current source and checked the printed output matches what the page claims, including the `halt_before` precedence behaviour and the resumed-run input handling.
- Checked that `.. _human-in-the-loop:` resolves from all three referring locations, and that the label sits after the page title to match `actions.rst` and `streaming-actions.rst`.

## Notes

Docs only, no source changes. The examples referenced (`email-assistant`, `conversational-rag/graph_db_example`) are quoted rather than modified.

Happy to shorten the page or move it if you'd rather this live under `docs/concepts/actions.rst` as a subsection instead of standing on its own.

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/concepts/actions.rst
+++ b/docs/concepts/actions.rst
@@ -322,6 +322,8 @@ We need to pass in `client` and `prompt` somehow. Here are the ways to do that:
 For instance, say you have a chatbot. The first step will likely declare the ``input`` parameter ``prompt`` --
 it will take that, process it, and put the result in state. The subsequent steps will read the result of that from state.
 
+To stop before an action that needs a person, then resume with ``inputs``, see :ref:`human in the loop <human-in-the-loop>`.
+
 ---------------
 Tagging Actions
 ---------------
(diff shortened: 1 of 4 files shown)


jernejfrank (approved):
Cool, thanks for adding this!

PR #894, opened 2026-08-18
Fix the Graph API example on the applications concepts page

Excelius-Wang:
The `Graph` / `Application` split example on the applications page cannot be copy-pasted: it raises a `SyntaxError` before it gets anywhere.

## Changes

Two lines in `docs/concepts/state-machine.rst`:

- `with_state(chat_history=[])` was missing its leading dot, so the builder chain broke mid-expression.
- The import read `from burr.core import ApplicationBuilder, default, expr`, which brings in two names the snippet never uses while leaving `GraphBuilder` undefined. Changed to `from burr.core import ApplicationBuilder, GraphBuilder`.

## How I tested this

Ran the corrected snippet with a `human_input` / `ai_response` action pair defined as in the surrounding docs. It builds the graph and the application without error. The original raises `SyntaxError` at the `with_state` line, and after fixing only that, `NameError: name 'GraphBuilder' is not defined`.

## Notes

Docs only, and deliberately separate from #893 so this one can go in on its own. Both branches touch `state-machine.rst` but not the same lines, so there is no conflict either way round.

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/concepts/state-machine.rst
+++ b/docs/concepts/state-machine.rst
@@ -171,7 +171,7 @@ in a web-server (create a graph once, application many times), import the graph 
 
 .. code-block:: python
 
-    from burr.core import ApplicationBuilder, default, expr
+    from burr.core import ApplicationBuilder, GraphBuilder
     graph = (
         GraphBuilder()
         .with_actions(human_input, ai_response)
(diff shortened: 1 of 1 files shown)


jernejfrank (approved):
Thanks!

PR #895, opened 2026-08-19
Fix async step persistence for synchronous actions

Eilen6316:
## Summary

When `Application.astep()` executes a synchronous action, it now forwards the action result and updated state returned by `_step()` to the async post-step hooks.

This prevents async persisters from saving the pre-step state as `completed` and ensures lifecycle hooks observe the actual result and state.

## Testing

- `tests/core/test_application.py`: 138 passed
- `tests/core` excluding optional Graphviz display tests: 390 passed
- `flake8 burr/core/application.py tests/core/test_application.py`

The two Graphviz display tests require the optional `graphviz` Python package, which is not installed in the test environment.

### DIFF
--- a/burr/core/application.py
+++ b/burr/core/application.py
@@ -1113,11 +1113,10 @@ class Application(Generic[ApplicationStateType]):
                     # but that's safer than assuming its OK to launch a thread
                     # TODO -- add an option/configuration to launch a thread (yikes, not super safe, but for a pure function
                     # which this is supposed to be its OK).
-                    # this delegates hooks to the synchronous version, so we'll call all of them as well
                     # In this case we allow the self._step to do input processing
-                    output = self._step(
-                        inputs=inputs, _run_hooks=False
-                    )  # Skip hooks as we already ran all of them/will run all of them in this function's finally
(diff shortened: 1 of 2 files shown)


burr/core/application.py
```diff
@@ -1115,25 +1115,29 @@ async def _astep(self, inputs: Optional[Dict[str, Any]], _run_hooks: bool = True
                     # which this is supposed to be its OK).
                     # this delegates hooks to the synchronous version, so we'll call all of them as well
                     # In this case we allow the self._step to do input processing
-                    return self._step(
+                    output = self._step(
                         inputs=inputs, _run_hooks=False
                     )  # Skip hooks as we already ran all of them/will run all of them in this function's finally
```
jernejfrank: This comment is out of place now, would be good to document inline here as well that we want to run async hooks for things like async persisters saving state of sync steps

jernejfrank (commented):
Great, just the one minor thing to keep the code clean. Looks good otherwise

Eilen6316:
> Could you also run pre-commit hooks please?

Sure! I’d be happy to take care of that.

jernejfrank (approved):

Issue #898, opened 2026-08-21
sdist tar.gz for releases needs its own LICENSE file


pjfanning:
One minor issue in the LICENSE file for the 'sdist' archive. It
mentions website files that are not in that archive. It looks like the
LICENSE is copied from the src archive and these files are in that.
It would be good to get this fixed for future RCs.

pjfanning:
fyi @jernejfrank

jernejfrank:
Resolved in #899

PR #899, opened 2026-08-21
Create sdist license

jernejfrank:
Addresses #898.

The root level license involves packages excluded in sdist; this adjust the release helper to create a temp dir and copying over the necessary artifacts for creating the sdist.

Flint only copies LICENSE so the rename happens during copy.


pjfanning (approved):
lgtm - thanks

